PR #10, opened 2023-03-28
fix: increase ETCD msg size limit further

vladklokun:
# What this PR changes?
- Remove gRPC message size limits
- Increase write request body max bytes to 1 GiB


matthyx (approved):
moarrrrrr!

PR #16, opened 2023-05-30
fix: reject Protobuf requests

vladklokun:
# What this PR changes?

Prior to this PR, the types that this APIServer declared did not support Protobuf, yet the default APIServer configuration advertised it as supported. This led to certain requests made in Protobuf format fail and could lead to error messages in the server logs and unexpected issues in the cluster. One of the issues we have found is not being able to delete namespaces when the APIServer was running (see https://github.com/kubernetes/kubernetes/issues/86666).

This PR introduces a default serializer that rejects the Protobuf content type, thus not advertizing support for it, making clients fall back to other supported content types.

### DIFF
--- a/pkg/apiserver/serializer_test.go
+++ b/pkg/apiserver/serializer_test.go
@@ -1,0 +1,89 @@
+package apiserver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type stubSerializer struct {
+	supportedTypes []runtime.SerializerInfo
+}
+
+func newStubSerializer(supportedTypes []runtime.SerializerInfo) *stubSerializer {
+	return &stubSerializer{supportedTypes: supportedTypes}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/apiserver/serializer_test.go
```diff
@@ -0,0 +1,77 @@
...
+			name: "Wrapping an original serializer should return original types",
+			originalContentTypes: []runtime.SerializerInfo{
+				{MediaType: runtime.ContentTypeProtobuf},
+				{MediaType: runtime.ContentTypeJSON},
+				{MediaType: runtime.ContentTypeYAML},
+			},
+			wantContentTypes: []runtime.SerializerInfo{
+				{MediaType: runtime.ContentTypeJSON},
+				{MediaType: runtime.ContentTypeYAML},
+			},
```
matthyx, line 54: do we want to add the case json+yaml -> json+yaml ?

vladklokun: Good catch! We do, please see the update.

matthyx (approved):

PR #27, opened 2023-06-16
add otel tracing and events

matthyx:


pkg/registry/file/watch_test.go
```diff
@@ -238,7 +237,7 @@ func TestFilesystemStoragePublishesToMatchingWatch(t *testing.T) {
 			var ttl uint64 = 0
 			var out runtime.Object
 			for key, object := range tc.inputObjects {
-				s.Create(ctx, key, object, out, ttl)
+				_ = s.Create(ctx, key, object, out, ttl)
```
vladklokun, line 240: question: is there any reason we assign to a throwaway variable?

pkg/registry/file/storage.go
```diff
@@ -24,21 +26,19 @@ import (
...
 type objState struct {
-	obj   runtime.Object
-	meta  *storage.ResponseMeta
-	rev   int64
-	data  []byte
-	stale bool
+	obj  runtime.Object
+	meta *storage.ResponseMeta
+	rev  int64
+	data []byte
```
vladklokun, line 38: question: is the `stale bool` variable no longer needed? Why? Was it used before?

main.go
```diff
@@ -17,14 +17,34 @@ limitations under the License.
...
+	"github.com/kubescape/storage/pkg/cmd/server"
+	"github.com/kubescape/storage/pkg/config"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/component-base/cli"
-	"github.com/kubescape/storage/pkg/cmd/server"
 )
 
 func main() {
+	ctx := context.Background()
+	c, err := config.LoadConfig("/etc/config")
```
vladklokun, line 34: question: this would mean that the storage would need to have this config file present?

How would this be mapped in the Helm chart? How would we ensure this config exists? Should we provide an example?

pkg/registry/file/watch_test.go
matthyx: golangci-lint

pkg/registry/file/storage.go
matthyx: I think it was never needed, golangci-lint was complaining.

main.go
matthyx: either we need the file present, or having some env set in order to know the customerID to enrich spans
https://github.com/kubescape/helm-charts/pull/201

pkg/registry/file/watch_test.go
vladklokun: :ok_hand:

vladklokun (approved):

PR #28, opened 2023-06-26
feat: add summary objects

vladklokun:
#  What this PR changes?

This commit:
- Introduces a new API type: SBOM Summary. The new type acts as a lightweight beacon that will be created whenever an SBOM is created, so the Storage APIServer can avoid loading entire large SBOMs into memory when listing and serving Watch requests.
- Ditto for new type Vulnerability Manifest Summary.
- Fixes a conflict with Makefile that was introduced when creating a `build` directory in the source tree.


dwertent (approved):

pkg/registry/softwarecomposition/vulnerabilitymanifest/etcd.go
```diff
@@ -52,5 +52,5 @@ func NewREST(scheme *runtime.Scheme, storageImpl storage.Interface, optsGetter g
 	if err := store.CompleteWithOptions(options); err != nil {
 		return nil, err
 	}
-	return &registry.REST{store}, nil
+	return &registry.REST{Store: store}, nil
```
matthyx, line 55: yay golint!

matthyx (approved):
much easier when looking at separate commits, thx!

pkg/registry/softwarecomposition/vulnerabilitymanifest/etcd.go
vladklokun: Kinda funny I found this with my LSP (`go vet`), since the linter did not even run on PR changes for me :)

PR #30, opened 2023-07-27
Kubescape workload scan types

amirmalka:
New types for saving Kubescape workload scan results (https://github.com/kubescape/kubescape/pull/1300)


pkg/apis/softwarecomposition/types.go
```diff
@@ -216,3 +216,109 @@ type VulnerabilityManifestSummaryList struct {
...
+	Info      string
+}
+
+type ScannedControlRule struct {
+	Name                  string
+	Status                RuleStatus
+	ControlConfigurations map[string][]string
+	Paths                 []RulePath
+	AppliedIgnoreRules    []string
+	RelatedResourcesIDs   []string // ?
```
amirmalka, line 268: @dwertent not sure about this field -  https://github.com/kubescape/opa-utils/blob/master/reporthandling/results/v1/resourcesresults/datastructures.go#L35

pkg/apis/softwarecomposition/types.go
```diff
@@ -216,3 +216,109 @@ type VulnerabilityManifestSummaryList struct {
...
+	Controls       map[string]ScannedControl
+	RelatedObjects []WorkloadScanRelatedObject
+}
+
+type ScannedControl struct {
+	ControlID string
+	Name      string
+	Severity  ControlSeverity
+	Status    ScannedControlStatus
+	Rules     []ScannedControlRule
```
vladklokun, line 248: question: storing rules inside objects is the only way we can proceed now, right?

The alternative would be storing these definitions somewhere inside the cluster, and we have decided we are not doing this at the moment. Is that correct?

vladklokun (commented):
Other than the question about rules and controls, this looks good to me. Now I can use it to implement the types in the storage.

pkg/apis/softwarecomposition/types.go
vladklokun: Resolved on our meeting: yes, we absolutely need this at the moment, and this field stays.

amirmalka:
@vladklokun so let's merge this :)

rcohencyberarmor (approved):

PR #37, opened 2023-08-18
Summary api objects

rcohencyberarmor:
support vuln summary object:
1. add new type VulnerabilitySummary and generate relevant files (new type doc: https://cyberarmor-io.atlassian.net/wiki/spaces/RD/pages/216727553/Summary+vuln+object+CRD+modification)
2. this type is not actually store in the storage, it is always regenerate for every K8S GET API 
3. all of other K8S API to this object will return error that claims that the current API used is not supported


pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+
+// add uni test
+func (s *VulnSummaryStorageImpl) Watch(ctx context.Context, key string, _ storage.ListOptions) (watch.Interface, error) {
+	_, span := otel.Tracer("").Start(ctx, "VulnSummaryStorageImpl.Watch")
+	span.SetAttributes(attribute.String("key", key))
+	defer span.End()
+	return nil, storage.NewMethodNotImplementError(key, "")
+}
+
+// add uni test
```
vladklokun, line 68: issue: this should not be in the code base

If we need to add unit test, the best course of action is to add it. If for some reason we cannot, we should add a `TODO(person_responsible)` with an explanation of what exactly we need to test and why we can’t do it now.

And having meaningful unit tests is mandatory before merging this PR.

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if _, exist := vulnManifestSummary.GetLabels()["kubescape.io/workload-namespace"]; !exist {
+		return nil, storage.NewKeyNotFoundError(path, 0)
+	}
+	return &vulnManifestSummary, nil
+}
+
+// add uni test
+func initNamespaceVulnSummary(key string) softwarecomposition.VulnerabilitySummary {
+	namespace := filepath.Base(key)
```
vladklokun, line 93: issue: namespace should not be computed inside the function, it should be a parameter

Essentially, this function is a constructor for VulnerabilitySummaries. That means, it should create an empty struct and populate it with necessary default values. It is not the job of the constructor to compute these values from whatever noise comes into it.

Instead, if we review what happens on a high level inside the storage, the flow is as follows:
- A user makes a request to K8s API
- K8s API forwards this request to our Storage APIServer
- Storage APIServer routes this to appropriate `REST` objects
- The `REST` object passes the `path` value to the `Storage` objects
- The `Storage` object starts handling the requested operation.
- The inner objects handle things based on some business logic we define.

Now, this business logic cares about creating a Vulnerability Summary for some scope, be it namespace or cluster. It does not care about the entire path parameter. And it should not even know about the concept of an element path. All it has to know is the name of the scope that it needs to create a Vulnerability Summary object for. 

Which is why this function should take a `namespace` string instead of computing it on its own. The caller should supply the namespace.

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func updateFullVulnSumm(fullVulnSumm *softwarecomposition.VulnerabilitySummary, vulnManifestSumm *softwarecomposition.VulnerabilityManifestSummary) {
+	updateSeverities(&fullVulnSumm.Spec.Severities, &vulnManifestSumm.Spec.Severities)
+	if vulnManifestSumm.Spec.Vulnerabilities.ImageVulnerabilitiesObj != (softwarecomposition.VulnerabilitiesObjScope{}) {
+		fullVulnSumm.Spec.WorkloadVulnerabilitiesObj = append(fullVulnSumm.Spec.WorkloadVulnerabilitiesObj, vulnManifestSumm.Spec.Vulnerabilities.ImageVulnerabilitiesObj)
+	}
+	if vulnManifestSumm.Spec.Vulnerabilities.WorkloadVulnerabilitiesObj != (softwarecomposition.VulnerabilitiesObjScope{}) {
+		fullVulnSumm.Spec.WorkloadVulnerabilitiesObj = append(fullVulnSumm.Spec.WorkloadVulnerabilitiesObj, vulnManifestSumm.Spec.Vulnerabilities.WorkloadVulnerabilitiesObj)
+	}
```
vladklokun, line 127: issue: it is hard to understand what this function does and why

What are we doing here? Are we merging manifest summaries? Which ones? _Why_ might we need to merge them?

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (s *VulnSummaryStorageImpl) getVulnManifestSummaryDirPath(key string) string {
+	return filepath.Join(s.root, key, "..", "..", "..", "vulnerabilitymanifestsummaries")
```
vladklokun, line 131: issue: function jumps up 3 levels of “path” without being tested

In general, the “key” being passed is not a file path. It is a concidence that we use it to store the file within the file storage implementation. But this isn’t the file storage. This is the Vulnerability Manifest storage, and it should have no business dealing with files or any file-related logic.

Instead, it makes more sense to re-evaluate what we need to build a summary of vulnerabilities. To me, it seems that we need at least the name of the scope that we are building for (be it namespace or cluster) and a way to fetch vulnerability summaries by a filter like namespace. Then, this function would not be needed.

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+		if !strings.HasSuffix(path, jsonExt) {
+			return nil
+		}
+		e := s.aggregateVulnSummaryOverNamespace(ctx, &fullVulnSumm, path)
+		if err != nil {
+			errs = append(errs, e)
+		}
+		return nil
+	})
```
vladklokun, line 165: issue: this code deals with low-level details of how vulnerability manifests get stored

This is what I was afraid of: this Vulnerability Storage object now dives into the details of how Vulnerability Manifests are stored.

Consider this: if we change the `file.StorageImpl` to change the way it names the files it stores, we would have to change this function. This breaks at least the single responsibility principle, will mess up our code base and slow us down.

This function should have no idea how vulnerability manifests are stored. Instead, it should have a way to call the Vulnerability Manifest storage to fetch Vulnerability Manifests by the parameters it needs: namespace or entire cluster.  Then, it can be completely oblivious to how the Vulnerability Manifest storage is implemented, tolerate any future changes and still do its job efficiently.

One way of implementing this would be to add Vulnerability Manifest storage as a parameter to the Vulnerability Summary storage. Then, the Vulnerability Summary storage can call the Vulnerability Manifest storage to retrieve Vulnerability manifests. In pseudo-code, this idea would look like:
```go
vmStore := file.StorageImpl()
kind := "VulnerabilityManifest"
vms := vmStore.GetByNamespace(kind, namespace)
summary := SummarizeVulnerabilities(vms)
return summary
```

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+func (s *VulnSummaryStorageImpl) Get(ctx context.Context, key string, opts storage.GetOptions, objPtr runtime.Object) error {
+	ctx, span := otel.Tracer("").Start(ctx, "VulnSummaryStorageImpl.Get")
+	span.SetAttributes(attribute.String("key", key))
+	defer span.End()
+
+	err := s.validateKeySupported(ctx, key)
+	if err != nil {
+		return err
+	}
+
```
vladklokun, line 196: thought: this is where we should extract the meaningful data from the key operation

We should extract the kind, namespace and name of the resource that we are requested to process here, and then pass it some business-level object that handles the business logic of building the summary.

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fs := afero.NewMemMapFs()
+			_ = fs.Mkdir(DefaultStorageRoot, 0755)
+			for _, f := range files {
+				_ = afero.WriteFile(fs, DefaultStorageRoot+f, []byte(""), 0644)
+			}
+			s := NewStorageImpl(fs, DefaultStorageRoot)
```
vladklokun, line 62: issue: this tests the `Count()` test for SBOMs and has no relation to the summary objects

We have no need for this, so it should be adapted to the actual implementation of the `Count()` object of this storage.

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fs := afero.NewReadOnlyFs(afero.NewMemMapFs())
+			s := NewVulnSummaryStorageImpl(fs, DefaultStorageRoot)
+			err := s.Create(context.TODO(), tt.args.key, tt.args.obj, tt.args.out, tt.args.in4)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
```
vladklokun, line 106: suggestion: we should test for the error that we want instead of some blanket error

How do you know that you got an error for the correct reason? How do you know the error you get is the one that you expect, and not some other error?

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fs := afero.NewReadOnlyFs(afero.NewMemMapFs())
+			s := NewVulnSummaryStorageImpl(fs, DefaultStorageRoot)
+			err := s.Delete(context.TODO(), tt.args.key, tt.args.obj, tt.args.precondition, tt.args.validateDeletionFunc, tt.args.cachedObj)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
```
vladklokun, line 139: suggestion: same as above

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+	s := VulnSummaryStorageImpl{
+		appFs: afero.NewMemMapFs(),
+		root:  "/data",
+	}
+
+	for _, td := range tests {
+		res := s.getVulnManifestSummaryDirPath(td.input)
+		assert.Equal(t, td.expected, res)
+	}
+}
```
vladklokun, line 166: issue: after the architectural changes requested above, this should not be tested

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+func updateSeverities(aggregatedSeverities *softwarecomposition.SeveritySummary, severities *softwarecomposition.SeveritySummary) {
+	updateVulnCounters(&aggregatedSeverities.Critical, &severities.Critical)
+	updateVulnCounters(&aggregatedSeverities.High, &severities.High)
+	updateVulnCounters(&aggregatedSeverities.Medium, &severities.Medium)
+	updateVulnCounters(&aggregatedSeverities.Low, &severities.Low)
+	updateVulnCounters(&aggregatedSeverities.Negligible, &severities.Negligible)
+	updateVulnCounters(&aggregatedSeverities.Unknown, &severities.Unknown)
+}
+
+func updateFullVulnSumm(fullVulnSumm *softwarecomposition.VulnerabilitySummary, vulnManifestSumm *softwarecomposition.VulnerabilityManifestSummary) {
```
vladklokun, line 120: thought: should these be struct methods for the appropriate structs?

Seeing all these functions take an API object of some kind made me wonder if they should be methods on corresponding structs. For example, the `updateFullVulnSumm` with its terse name seem like a `VulnerabilitySummary.Merge(other)` method to me. Same for the updating the vulnerability counters. Wouldn’t they make more sense as `VulnerabilityCounters.Add(other)` method?

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+			assert.Equal(t, tt.fullVulnSumm.Spec.Severities.Negligible.All, tt.expectedFullVulnSumm.Spec.Severities.Negligible.All)
+			assert.Equal(t, tt.fullVulnSumm.Spec.Severities.Negligible.Relevant, tt.expectedFullVulnSumm.Spec.Severities.Negligible.Relevant)
+
+			assert.Equal(t, tt.fullVulnSumm.Spec.Severities.Unknown.All, tt.expectedFullVulnSumm.Spec.Severities.Unknown.All)
+			assert.Equal(t, tt.fullVulnSumm.Spec.Severities.Unknown.Relevant, tt.expectedFullVulnSumm.Spec.Severities.Unknown.Relevant)
+
+			assert.Equal(t, len(tt.fullVulnSumm.Spec.WorkloadVulnerabilitiesObj), len(tt.expectedFullVulnSumm.Spec.WorkloadVulnerabilitiesObj))
+			for i := range tt.fullVulnSumm.Spec.WorkloadVulnerabilitiesObj {
+				assert.Equal(t, tt.fullVulnSumm.Spec.WorkloadVulnerabilitiesObj[i], tt.expectedFullVulnSumm.Spec.WorkloadVulnerabilitiesObj[i])
+			}
```
vladklokun, line 317: question: why make all these asserts separately when you provide an entire expected object as a test input?

Wouldn’t it be more convenient to just compare the objects like:
```
got := FuncBeingTested()
assert.Equal(t, tc.expectedFullVulnSumm, got)
```

pkg/registry/file/vulnsummarystorage_test.go
```diff
@@ -0,0 +1,461 @@
...
+			assert.Equal(t, tt.aggregatedVulnSeverities.Medium.Relevant, tt.expectedRelevantMedium)
+
+			assert.Equal(t, tt.aggregatedVulnSeverities.Low.All, tt.expectedAllLow)
+			assert.Equal(t, tt.aggregatedVulnSeverities.Low.Relevant, tt.expectedRelevantLow)
+
+			assert.Equal(t, tt.aggregatedVulnSeverities.Negligible.All, tt.expectedAllNegligible)
+			assert.Equal(t, tt.aggregatedVulnSeverities.Negligible.Relevant, tt.expectedRelevantNegligible)
+
+			assert.Equal(t, tt.aggregatedVulnSeverities.Unknown.All, tt.expectedAllUnknown)
+			assert.Equal(t, tt.aggregatedVulnSeverities.Unknown.Relevant, tt.expectedRelevantUnknown)
```
vladklokun, line 427: question: same as above

vladklokun (changes requested):
All in all, I think we need to overhaul this approach from the ground up. We should consider storage as something that deals with how to store and present objects, not with how to build them. A builder of the objects should be a separate entity, and out of scope for the storage context.

This leads to a major issue I see in the implementation: it has to know and keep up with the details of file storage that it should not know about.

Instead, our approach should be focused on re-using the existing storage. This would allow us to isolate the object fetching logic in only one place instead of having to carry it all over the other storage implementations.

Another big issue is these little private functions that perform calculations on API objects. To me, they seem like they should be one of the two:
- Methods on the API objects themselves. I feel like  `Summary.Merge()` and `Counters.Add()` are natural fits for the functions I commented on.
- Methods on some other builder object / package / entity that is responsible for building the Summary objects, but not storing or presenting them.

pkg/registry/file/vulnsummarystorage.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Versioner Returns Versioner associated with this interface.
+func (s *VulnSummaryStorageImpl) Versioner() storage.Versioner {
+	return s.versioner
+}
+
+func (s *VulnSummaryStorageImpl) Create(ctx context.Context, key string, obj, out runtime.Object, _ uint64) error {
+	_, span := otel.Tracer("").Start(ctx, "VulnSummaryStorageImpl.Create")
+	span.SetAttributes(attribute.String("key", key))
```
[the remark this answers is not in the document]
rcohencyberarmor: I'll remove it

pkg/registry/file/vulnsummarystorage.go
rcohencyberarmor: please take a look on the docs reference

rcohencyberarmor:
@vladklokun for now we are not going to support cluster scope, just namespace scope.

PR #46, opened 2023-10-11
feat: WIP add control etc crds

XDRAGON2002:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces several new Custom Resource Definitions (CRDs) and their related functionalities:
- Control: Represents a control entity in the software composition schema.
- Rule: Represents a rule entity in the software composition schema.
- Framework: Represents a framework entity in the software composition schema.
- Exception: Represents an exception entity in the software composition schema.

For each of these CRDs, the PR includes the necessary conversion functions for the software composition scheme. It also updates the clientset, listers, and informers to handle these new CRDs.

___
## PR Main Files Walkthrough:

- `pkg/apis/softwarecomposition/v1beta1/zz_generated.conversion.go`: Added conversion functions for the new CRDs (Control, Rule, Framework, Exception).
- `pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/controlconfiguration.go`: Updated the clientset to handle the new ControlConfiguration CRD.
- `pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/framework.go`: Updated the clientset to handle the new Framework CRD.
- `pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/exception.go`: Updated the clientset to handle the new Exception CRD.
- `pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/rule.go`: Updated the clientset to handle the new Rule CRD.
- `pkg/apis/softwarecomposition/v1beta1/types.go`: Added definitions for the new CRDs (Control, Rule, Framework, Exception).
- `pkg/generated/informers/externalversions/softwarecomposition/v1beta1/controlconfiguration.go`: Updated the informers to handle the new ControlConfiguration CRD.
- `pkg/generated/informers/externalversions/softwarecomposition/v1beta1/exception.go`: Updated the informers to handle the new Exception CRD.
- `pkg/generated/informers/externalversions/softwarecomposition/v1beta1/framework.go`: Updated the informers to handle the new Framework CRD.
- `pkg/generated/informers/externalversions/softwarecomposition/v1beta1/rule.go`: Updated the informers to handle the new Rule CRD.
</details>


XDRAGON2002:
@vladklokun as discussed, I've been working on this for the past week, right now I'm stuck a bit as I just can't seem to be able to successfully run the `codegen` script `hack/update-codegen.sh`.

The script always seems to error out/give out warnings? Not even sure what they are, every time I run it but can't seem to generate the required deepcopy methods, also why they are not present in the PR.

I've looked for solutions online but can't seem to get it working still, any help regarding this would be extremely helpful! Thanks!

vladklokun:
> The script always seems to error out/give out warnings? Not even sure what they are, every time I run it but can't seem to generate the required deepcopy methods, also why they are not present in the PR.

Yes, the script always generates warnings when running, even after successfully generating deep copy methods, converters and OpenAPI definitions. That’s normal.

You can tell your script ran okay-ish when it generates all the methods you need.

Regarding the missing methods, I know it might sound exotic, but where have you cloned your project? There is a catch with `hack/update_codegen.sh` that it outputs its definitions in the `$HOME/github.com/kubescape/storage/pkg` directory, provided you installed the dependencies with the `go mod vendor` method. To you have generated files in this directory, by chance?

XDRAGON2002:
@vladklokun finally got the codegen working, could you maybe take a look and let me know if I've missed something before I go ahead and resolve the merge conflicts?

Thanks!

vladklokun:
Hi! I finally had a chance to look at the PR. Thank you for your patience.

Type definitions and wiring look good to me.

However, we just merged a PR that includes changes to codegen. Please rebase and try to fix conflicts with the base branch and try to verify that the Custom Resources work. A good start to manually test this would be to create example Custom Resources in the `artifacts/{rule,framework,control,...}` directory, deploy your APIServer in a cluster using the manifests in the `artifacts/example` directory and then to create the example Custom Resources you put in the `artifacts` subdirectories.

XDRAGON2002:
Sure!

Will do that and get back to you on this.

Thanks!

XDRAGON2002:
@vladklokun have made the required changes! Thanks!

XDRAGON2002:
@dwertent Completely forgot about that! Thanks! Done!

XDRAGON2002:
This is weird, the tests still failed due to the `go.mod/sum` dependency issue, I already did push the updated ones though.

go.mod
```diff
@@ -4,6 +4,8 @@ module github.com/kubescape/storage
 
 go 1.21.3
 
+toolchain go1.21.5
+
```
vladklokun, line 8: question: why the differing toolchain, and in the patch versions, no less?

Why do we specifically need the `toolchain go1.21.5`? What’s the difference between the `1.21.3` and `1.21.5` that we explicitly require the latter? Is there something that prevents us from bumping the Go version to 1.21.5?

go.mod
```diff
@@ -17,7 +19,7 @@ require (
 	go.opentelemetry.io/otel v1.19.0
 	golang.org/x/exp v0.0.0-20231006140011-7918f672742d
 	k8s.io/api v0.26.2
-	k8s.io/apimachinery v0.26.2
+	k8s.io/apimachinery v0.28.2
 	k8s.io/apiserver v0.26.2
 	k8s.io/client-go v0.26.2
```
vladklokun, line 24: question: why is this bump needed?

Bumping `apimachinery` usually requires bumping `client-go`, too. Why do we need to use `apimachinery v0.28.2` version for this PR? And why don’t we bump the `go-client` with it, too?

JFYI, we might need to test this change with components that import the Storage Client, because what often happens when bumping `apimachinery`, especially without `k8s.io/client-go` is that clients start sending malformed requests and misbehave. So if we bump `k8s.io/apimachinery`, we need to bump `k8s.io/client-go`. And we also need to bump the Storage client on every single client that imports the Storage client anyway.

But then we will need to manually test the clients that import the Storage client, because `apimachinery` bumps can be breaking.

pkg/apis/softwarecomposition/types.go
```diff
@@ -596,3 +597,104 @@ type SBOMSyftFilteredList struct {
...
+	Payload unstructured.Unstructured
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ControlConfigurationList struct {
+	metav1.TypeMeta
+	metav1.ListMeta
+	Items []ControlConfiguration
+}
```
vladklokun, line 700: chore: fix line endings in PR

Please see the sign after this line. It indicates that the files have improper line endings: they don’t have the terminating `\n` characters. I believe it’s because you’re using Visual Studio Code or some other IDE that terminates the files with Windows line endings.

Please refer to this issue: https://github.com/microsoft/vscode/issues/141169

vladklokun (commented):
Other than formatting and the `apimachinery` bump, LGTM. But before anyone approves, we need to make sure the `apimachinery` bump actually works with all of our clients importing the Storage and doesn’t break anything.

go.mod
XDRAGON2002: I hadn't explicitly updated this tbh.

I just ran `go mod tidy` and it put this in (which on a second thought wasn't there earlier).

I'm currently on version 1.21.5

go.mod
XDRAGON2002: Same reason as above, hadn't made this change explicitly, it got pulled when I moved from 1.20 to 1.21

pkg/apis/softwarecomposition/types.go
XDRAGON2002: I'll fix this,

also the fact that VSC does this is just concerning, I'd at least hope to have my tooling follow standards :(

go.mod
vladklokun: Okay, so it should be safe to use:

```suggestion
go 1.21.5

toolchain go1.21.5

```

Please make the change and test it out.

pkg/apis/softwarecomposition/types.go
vladklokun: Yeah, that’s a common problem. Nonetheless, we could always add a Git pre-commit hook to fix the line endings automatically or a linter to complain about it. But as you can imagine, this is very low on our priorities list :)

go.mod
vladklokun: I see. If I were you, I would double-check that `go mod tidy` always bumps `apimachinery`. If it does, then I would bump `client-go` to `v0.28.2`, and try it out with the bumped modules. If all works, then yay. If not, then we will have to pin the apimachinery code.

XDRAGON2002:
@vladklokun @dwertent I'm pretty sure the tests will fail again, I can't seem to figure out why it is so.

Any help would be much appreciated!

PR #50, opened 2023-10-12
add dependabot config

matthyx:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces a configuration file for Dependabot, a tool that helps to keep dependencies up-to-date. The configuration file specifies that updates should be checked on a weekly basis. However, the package ecosystem and the directory of package manifests are not specified and need to be filled in.

___
## PR Main Files Walkthrough:

`.github/dependabot.yaml`: A new configuration file for Dependabot has been added. It is set to check for updates on a weekly basis. The package ecosystem and the directory of package manifests are left blank and need to be specified.
</details>

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"


vladklokun:
@matthyx @dwertent

Before we jump into having dependabot spam the PRs with bumping updates, I want us to consider the updating strategy for storage. What are we trying to achieve here?

My main concern is that the default workflow the Kubernetes SIG suggests for Aggregated APIServers forking from the sample repository is rebasing onto the parent repo every once in a while. Not bumping dependencies on your own. We never did this yet, because we don’t fully understand the implications like how much merge conflicts we would have to solve when rebasing onto the parent repository.

If we enable dependabot and commit to bumping images as it suggests, how does this affect our workflow? Does it make it more complicated to rebase onto the parent repo and keep the dependencies updated?

Let’s maybe define our dependencies strategy before enabling dependabot?

matthyx:
@vladklokun you're right

PR #56, opened 2023-10-18
Network neighbors

Daniel-GrunbergerCA:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces a new Custom Resource Definition (CRD) called `NetworkNeighbors`. This CRD will be used by the node-agent to save network communication information. The PR includes the addition of new types, registry, and merge strategy for the `NetworkNeighbors` CRD. It also includes the necessary changes in the generated files for the new CRD.

___
## PR Main Files Walkthrough:

`pkg/apis/softwarecomposition/v1beta1/network_types.go`: The new types related to the `NetworkNeighbors` CRD are defined here. This includes `NetworkNeighbors`, `NetworkNeighborsList`, `NetworkNeighborsSpec`, `IngressEntry`, `EgressEntry`, and `NetworkPort`.
`pkg/registry/softwarecomposition/networkneighbors/strategy.go`: The strategy for the `NetworkNeighbors` CRD is defined here. This includes the strategy for creating, updating, and resetting the `NetworkNeighbors` object.
`pkg/registry/softwarecomposition/networkneighbors/etcd.go`: The storage and registry settings for the `NetworkNeighbors` CRD are defined here.
`pkg/generated/openapi/zz_generated.openapi.go`: The OpenAPI definitions for the new types are added here.
`pkg/apis/softwarecomposition/v1beta1/zz_generated.deepcopy.go`: The auto-generated deep copy functions for the new types are added here.
`pkg/generated/informers/externalversions/softwarecomposition/v1beta1/interface.go`: The informer for the `NetworkNeighbors` CRD is added here.
`pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/fake/fake_softwarecomposition_client.go`: The fake client for the `NetworkNeighbors` CRD for testing purposes is added here.
`pkg/generated/clientset/versioned/typed/softwarecomposition/v1beta1/networkneighbors.go`: The client for the `NetworkNeighbors` CRD is added here.
`pkg/apiserver/apiserver.go`: The `NetworkNeighbors` CRD is registered in the API server.
`artifacts/networkneighborses/01-example.yaml`: An example YAML file for the `NetworkNeighbors` CRD is added here.
</details>

___
## User Description:
This PR adds a new CRD - `NetworkNeighbors`. This CRD will be used by the node-agent to save network communication information.


artifacts/networkneighborses/01-example.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: spdx.softwarecomposition.kubescape.io/v1beta1
+kind: NetworkNeighbors
+metadata:
+  name: deployment-nginx
+  labels:
+      "kubescape.io/workload-api-group": "spdx.softwarecomposition.kubescape.io"
+      "kubescape.io/workload-api-version" : "v1beta1"
```
vladklokun, line 7: chore: fix extra space

```suggestion
      "kubescape.io/workload-api-version": "v1beta1"
```

artifacts/networkneighborses/01-example.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  egress:
+  - type: external
+    identifier: bla
+    ipAddress: 123.5.2.3
+    dns: stripe.com 
+    ports:
+    - name: TCP-5978
+      protocol: TCP
+      port: 5978
```
vladklokun, line 38: chore: fix missing newline

The symbol at the bottom shows that there is a missing proper newline for the file. We should have proper unified line endings in the repo.

pkg/registry/softwarecomposition/networkneighbors/etcd.go
```diff
@@ -0,0 +1,41 @@
...
+	strategy := NewStrategy(scheme)
+
+	dryRunnableStorage := genericregistry.DryRunnableStorage{Codec: nil, Storage: storageImpl}
+
+	store := &genericregistry.Store{
+		NewFunc:                   func() runtime.Object { return &softwarecomposition.NetworkNeighbors{} },
+		NewListFunc:               func() runtime.Object { return &softwarecomposition.NetworkNeighborsList{} },
+		PredicateFunc:             MatachNetworkNeighbor,
+		DefaultQualifiedResource:  softwarecomposition.Resource("networkneighborses"),
+		SingularQualifiedResource: softwarecomposition.Resource("networkneighbors"),
```
vladklokun, line 24: thought: can we use only plural for resources instead of `networkneighborses`?

That sure sounds ugly and is not proper grammar. What if we just provide the `DefaultQualifiedResource` to be `networkneighbors` and refer to it like that?

vladklokun (changes requested):

pkg/apis/softwarecomposition/v1beta1/network_types.go
```diff
@@ -0,0 +1,67 @@
...
+// NetworkNeighbors represents a list of network communications for a specific workload.
+type NetworkNeighbors struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	Spec NetworkNeighborsSpec `json:"spec"`
+}
+
+type NetworkNeighborsSpec struct {
+	*metav1.LabelSelector
```
vladklokun, line 41: issue: missing the JSON struct tag

In order to properly embed a type into a JSON type definition, you need to provide the `json:",inline"` struct tag. Then the fields will be properly propagated to the embedding type.


```suggestion
	*metav1.LabelSelector `json:",inline"`
```

vladklokun: And why embed it as a pointer for the `*metav1.LabelSelector`? I think this should be:

```suggestion
	metav1.LabelSelector `json:",inline"`
```

pkg/apis/softwarecomposition/network_types.go
```diff
@@ -0,0 +1,61 @@
...
+	Ports             []NetworkPort
+	PodSelector       *metav1.LabelSelector
+	NamespaceSelector *metav1.LabelSelector
+	IPAddress         string
+}
+
+type NetworkPort struct {
+	Name     string // protocol-port
+	Protocol Protocol
+	Port     uint16
```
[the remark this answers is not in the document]
Daniel-GrunbergerCA: When should `uint` be used if not for positive values?

pkg/registry/softwarecomposition/networkneighbors/etcd.go
Daniel-GrunbergerCA: I don't get the suggestions. How should a single resource be called?

pkg/apis/softwarecomposition/network_types.go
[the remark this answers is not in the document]
vladklokun: I’m talking more about size of the type, not its sign. But now that I look at how [Deployments](https://github.com/kubernetes/api/blob/fe172d7dd4b6501923938705cfb69ee0b907c32d/apps/v1/types.go#L376) and [Flow Control](https://github.com/kubernetes/api/blob/fe172d7dd4b6501923938705cfb69ee0b907c32d/flowcontrol/v1beta3/types.go#L155) are implemented. It seems that it’s a convention to specify the size, but avoid `uint` even for non-negative values.

I believe this is because of serialization, so that overflowing values don’t get silently cast to `uint`, and instead get handled at the validator. For example, if you pass `replicas: -1`, a server will be able to deserialize it properly, then validate the value and give you a better, meaningful error.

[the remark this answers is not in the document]
Daniel-GrunbergerCA: I see. I actually made it based on the type that `inspektor gadget` uses at the event they create, but you are right

pkg/apis/softwarecomposition/network_types.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type NetworkNeighborsSpec struct {
+	metav1.LabelSelector // The labels which are inside spec.selector in the parent workload.
+	Ingress              []NetworkEntry
+	Egress               []NetworkEntry
+}
+
+// IngressEntry represents a single network communication.
+type NetworkEntry struct {
```
vladklokun, line 47: chore: we must make the naming match across comments and type definitions

Also, I have trouble understanding the comment. What’s a “single network communication”? And Network Entry seems too generic. It seems that we are describing a single network neighbor.

If we call our top-level resources Network Neighbors, wouldn’t it make sense that the entries in the `Spec.Ingress` list would have the type `NetworkNeighbor` singular? It seems to be less confusing, in my opinion.

If you do change the name, please keep in mind you have to change them in both top-level and `v1beta1` directories and also re-generate the code.

pkg/apis/softwarecomposition/network_types.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type NetworkPort struct {
+	// Name is an artificial identifier of the network port. We use it for merging keys with Strategic Merge Patch.
+	// Format is `{protocol}-{port}`.
+	//
+	// Example: tcp-6881
+	Name     string // protocol-port
+	Protocol Protocol
+	Port     *int32
```
vladklokun, line 64: question: why pointer to `int32`?

If you referenced what they do in K8s, the use a pointer to communicate that the value is optional. Is `Port` an optional value here?

pkg/apiserver/apiserver.go
```diff
@@ -166,6 +167,7 @@ func (c completedConfig) New() (*WardleServer, error) {
 	v1beta1storage["applicationprofilesummaries"] = sbomregistry.RESTInPeace(applicationprofilesummary.NewREST(Scheme, storageImpl, c.GenericConfig.RESTOptionsGetter))
 	v1beta1storage["applicationactivities"] = sbomregistry.RESTInPeace(applicationactivity.NewREST(Scheme, storageImpl, c.GenericConfig.RESTOptionsGetter))
 
+	v1beta1storage["networkneighborses"] = sbomregistry.RESTInPeace(networkneighbors.NewREST(Scheme, storageImpl, c.GenericConfig.RESTOptionsGetter))
```
vladklokun, line 170: chore: if using only `networkneighbors` as a resource qualifier worked, we have to use `networkneighbors` here, too

pkg/registry/softwarecomposition/networkneighbors/strategy.go
```diff
@@ -0,0 +1,121 @@
...
+// NewStrategy creates and returns a networkNeighborsStrategy instance
+func NewStrategy(typer runtime.ObjectTyper) networkNeighborsStrategy {
+	return networkNeighborsStrategy{typer, names.SimpleNameGenerator}
+}
+
+// GetAttrs returns labels.Set, fields.Set, and error in case the given runtime.Object is not a Flunder
+func GetAttrs(obj runtime.Object) (labels.Set, fields.Set, error) {
+	apiserver, ok := obj.(*softwarecomposition.NetworkNeighbors)
+	if !ok {
+		return nil, nil, fmt.Errorf("given object is not a Flunder")
```
vladklokun, line 26:
```suggestion
		return nil, nil, fmt.Errorf("given object is not a NetworkNeighbors")
```

pkg/registry/softwarecomposition/networkneighbors/strategy.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return validatePorts(networkNeighbors)
+}
+
+// WarningsOnUpdate returns warnings for the given update.
+func (networkNeighborsStrategy) WarningsOnUpdate(ctx context.Context, obj, old runtime.Object) []string {
+	return nil
+}
+
+func validatePorts(networkNeighbors *softwarecomposition.NetworkNeighbors) field.ErrorList {
```
vladklokun, line 94: question: are we skipping tests for this validator?

We should not cut corners when adding a validator.

pkg/registry/softwarecomposition/networkneighbors/strategy.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, ingress := range networkNeighbors.Spec.Ingress {
+		for _, networkPort := range ingress.Ports {
+			if *networkPort.Port < 0 || *networkPort.Port > 65535 {
+				return field.ErrorList{field.Invalid(field.NewPath("spec").Child("ingress").Child("ports"), *networkPort.Port, "port must be in range 0-65535")}
+			}
+
+			expectedPortName := fmt.Sprintf("%s-%d", networkPort.Protocol, *networkPort.Port)
+			if networkPort.Name != expectedPortName {
+				return field.ErrorList{field.Invalid(field.NewPath("spec").Child("ingress").Child("ports"), *networkPort.Port, fmt.Sprintf("port name must be in format {protocol}-{port}, expected name: %s", expectedPortName))}
+			}
```
vladklokun, line 104: issue: these two logical validations are not separated in code

These should be separate validators. Something like `validatePortNumber` and `validatePortName`. Take a look how the original validators are structured in `pkg/apis/softwarecomposition/validation/validation.go`. They traverse the objects from the `Object` to `ObjectSpec`. Whenever you go down a level to a new object type, it should be a new validator. In this case:
- `ValidateNetworkNeighbors(nn *softwarecomposition.NetworkNeighbors, fldPath *field.Path)`
- `ValidateNetworkNeighborsSpec(nns *softwarecomposition.NetworkNeighborsSpec, fldPath *field.Path)`
- `ValidateNetworkNeighbor(nn *softwarecomposition.NetworkNeighbor, fldPath *field.Path)`
- `ValidatePort(p *softwarecomposition.Port, fldPath *field.Path)`
etc.

Since these validators would take a field path as an argument, it should also eliminate the problem of duplicating the same code when you validate the `NetworkEntry` / `NetworkNeighbor`. Since you could just pass a `fieldPath.Child("ingress")` or `fieldPath.Child("egress")` depending on what you are passing for validation.

pkg/registry/softwarecomposition/networkneighbors/strategy.go
```diff
@@ -0,0 +1,121 @@
...
+			if *networkPort.Port < 0 || *networkPort.Port > 65535 {
+				return field.ErrorList{field.Invalid(field.NewPath("spec").Child("egress").Child("ports"), *networkPort.Port, "port must be in range 0-65535")}
+			}
+
+			expectedPortName := fmt.Sprintf("%s-%d", networkPort.Protocol, *networkPort.Port)
+			if networkPort.Name != expectedPortName {
+				return field.ErrorList{field.Invalid(field.NewPath("spec").Child("egress").Child("ports"), *networkPort.Port, fmt.Sprintf("port name must be in format {protocol}-{port}, expected name: %s", expectedPortName))}
+			}
+		}
+	}
```
vladklokun, line 119: suggestion: ...and once you split the validators as above, you wouldn’t need to duplicate the code

pkg/apis/softwarecomposition/network_types.go
vladklokun: suggestion: how about we call this `NetworkNeighbor`, singular?

I think this will be more immediately obvious than calling it a generic `NetworkEntry`.

vladklokun (changes requested):

pkg/apis/softwarecomposition/network_types.go
Daniel-GrunbergerCA: this differentiates between an empty port or a port set to zero

pkg/apis/softwarecomposition/validation/valdiation_test.go
```diff
@@ -0,0 +1,124 @@
...
+								{
+									Port:     80,
+									Name:     "UDP-80",
+									Protocol: "UDP",
+								},
+							},
+						},
+					},
+				},
+			},
```
vladklokun, line 34: issue: validators return an empty error list on success, not nil

So we should expect an empty list. That will also let you greatly simplify your test assertion. We can also improve on naming our test case:
```suggestion
			name: "valid network neighbors should return an empty error list",
			networkNeighbors: softwarecomposition.NetworkNeighbors{
				Spec: softwarecomposition.NetworkNeighborsSpec{
					Ingress: []softwarecomposition.NetworkNeighbor{
						{
							Identifier: "test",
							Ports: []softwarecomposition.NetworkPort{
								{
									Port:     80,
									Name:     "UDP-80",
									Protocol: "UDP",
								},
							},
						},
					},
				},
			},
			expectedErrors: field.ErrorList{},
```

pkg/apis/softwarecomposition/validation/valdiation_test.go
```diff
@@ -0,0 +1,124 @@
...
+			errorsFound := 0
+			for _, actualError := range actualErrors {
+
+				for _, expectedError := range test.expectedErrors {
+					if actualError.Error() == expectedError.Error() {
+						errorsFound += 1
+					}
+				}
+			}
+			assert.Equal(t, len(test.expectedErrors), errorsFound)
```
vladklokun, line 121: suggestion: with our fix to the expected errors, we can simplify the assertion

```suggestion
			assert.Equal(t, test.expectedErrors, actualErrors)
```

If you need to compare the items regardless of order, try using `assert.ElementsMatch`.

vladklokun (changes requested):
Nice! Now the validators look good to me. Some very small improvements and we should be good to go :)

pkg/apis/softwarecomposition/network_types.go
vladklokun: Makes sense. Just to re-iterate our conversation in Slack, if we have it as a pointer, as you originally suggested, we can give a better error message when a client tries to create a NetworkNeighbor without a port value specified. I think that’s the way to do it.

It’s your call on whether to have `Port` as a pointer.

pkg/apis/softwarecomposition/validation/valdiation_test.go
```diff
@@ -0,0 +1,115 @@
...
+				Spec: softwarecomposition.NetworkNeighborsSpec{
+					Ingress: []softwarecomposition.NetworkNeighbor{
+						{
+							Identifier: "test",
+							Ports: []softwarecomposition.NetworkPort{
+								{
+									Port:     &port80,
+									Name:     "UDP-80",
+									Protocol: "UDP",
+								},
```
vladklokun, line 32: suggestion(non-blocking): use the pointer utils in K8s

K8s infrastructure provides utility functions for working with different kinds of pointers. We have the package as a dependency anyway, so we should use it.

```suggestion
	"k8s.io/apimachinery/pkg/util/validation/field"
	"k8s.io/utils/pointer"
)

func TestValidateNetworkNeighbors(t *testing.T) {
	tests := []struct {
		name             string
		port             int32
		networkNeighbors softwarecomposition.NetworkNeighbors
		expectedErrors   field.ErrorList
	}{
		{
			name: "valid",
			networkNeighbors: softwarecomposition.NetworkNeighbors{
				Spec: softwarecomposition.NetworkNeighborsSpec{
					Ingress: []softwarecomposition.NetworkNeighbor{
						{
							Identifier: "test",
							Ports: []softwarecomposition.NetworkPort{
								{
									Port:     pointer.Int32(80),
									Name:     "UDP-80",
									Protocol: "UDP",
								},
```

vladklokun (dismissed):
LGTM!

I want to re-iterate that the suggestion regarding the pointers is non-blocking, but I would appreciate if you made the change.

vladklokun:
Oh, and please make sure to rebase on main and squash the commits into something with a sensible commit message. I would suggest:
```
feat: add and expose the NetworkNeighbors custom resource
```

We should keep our commit history clear for checking at a glance. Having 10 commits named along the lines of "add types", "pr fixes" and a merge from main would pollute it.

vladklokun (approved):

pkg/apis/softwarecomposition/validation/valdiation_test.go
```diff
@@ -0,0 +1,114 @@
+package validation
```
vladklokun, line 1: typo: file is named `valdiation_test.go`

Should be `validation_test.go`.

vladklokun (changes requested):

PR #74, opened 2023-12-14
merge ports

Daniel-GrunbergerCA:
Merge IP addresses by ports in generated NP


pkg/apis/softwarecomposition/networkpolicy/networkpolicy.go
```diff
@@ -81,11 +82,151 @@ func GenerateNetworkPolicy(networkNeighbors softwarecomposition.NetworkNeighbors
...
+	networkPolicy.Spec.Egress = mergeEgressRulesByPorts(networkPolicy.Spec.Egress)
+
+	networkPolicy.Spec.Ingress = mergeIngressRulesByPorts(networkPolicy.Spec.Ingress)
+
 	generatedNetworkPolicy.Spec = networkPolicy
 
 	return generatedNetworkPolicy, nil
 }
 
+func mergeIngressRulesByPorts(rules []softwarecomposition.NetworkPolicyIngressRule) []softwarecomposition.NetworkPolicyIngressRule {
```
matthyx, line 94: you cannot define a comparable interface for rules and use a [set](https://github.com/deckarep/golang-set) for that?

Daniel-GrunbergerCA: I didn't use since we don't have duplications where both ports and IP are equal, and also we need to convert it to list anyways (since this is the struct in the NP)

dwertent (approved):

PR #82, opened 2024-01-02
adding routines for better performance of NP

yuleib:
adding routines for better performance of NP


dwertent:
We see the generating is very fast, is there is no need to add routines

PR #101, opened 2024-03-05
refactor locks to prevent race on get list

matthyx:


pkg/registry/file/storage.go
```diff
@@ -577,8 +565,6 @@ func (s *StorageImpl) GetByNamespace(ctx context.Context, apiVersion, kind, name
 	span.SetAttributes(attribute.String("apiVersion", apiVersion), attribute.String("kind", kind), attribute.String("namespace", namespace))
 	defer span.End()
 	_, spanLock := otel.Tracer("").Start(ctx, "waiting for lock")
```
amirmalka, line 567: also here

amirmalka (approved):

PR #125, opened 2024-07-14
Fix watcher data race

ttimonen:
Fix the watch data race

Also, simplify the underlying logic to lure out what is actually going on here.


matthyx (approved):
thanks @ttimonen !

PR #127, opened 2024-07-19
Improvements

matthyx:


pkg/cmd/server/start.go
```diff
@@ -180,6 +183,26 @@ func (o WardleServerOptions) RunWardleServer(stopCh <-chan struct{}) error {
...
+		memory.AdapterWithAlgorithm(memory.LRU),
+		memory.AdapterWithStorageCapacity(500000000), // 500MB
+	)
+	if err != nil {
+		logger.L().Fatal("failed to create memcached adapter", helpers.Error(err))
+	}
+
+	cacheClient, err := cache.NewClient(
+		cache.ClientWithAdapter(memcached),
+		cache.ClientWithTTL(10*time.Minute),
```
matthyx, line 197: obviously this needs to be set lower, or we should check how to invalidate cache

slashben: Yes indeed

slashben (commented):
I have looked into the `http-cache` package. My main concern is the lack of cache invalidation. I don't see it invalidates cache items on the `PUSH`/`PUT`/`UPDATE`/`DELETE` methods. Either we fork and implement this, or need to find another solution because this implementation leads to inconsistency.

PR #140, opened 2024-08-28
fix NN update strategy based on the AP one

matthyx:


pkg/registry/softwarecomposition/networkneighborhood/strategy.go
```diff
@@ -62,9 +64,22 @@ func (NetworkNeighborhoodStrategy) PrepareForUpdate(ctx context.Context, obj, ol
 	newAP := obj.(*softwarecomposition.NetworkNeighborhood)
 	oldAP := old.(*softwarecomposition.NetworkNeighborhood)
 
+	// if we have an network neighborhood that is marked as complete and completed, we do not allow any updates
+	if oldAP.Annotations[helpers.CompletionMetadataKey] == helpers.Complete && oldAP.Annotations[helpers.StatusMetadataKey] == helpers.Completed {
```
amitschendel, line 68: Why it's called `oldAP` and not `oldNN`?
Same for the rest.

matthyx: it was like that... at least it simplifies the diffs

matthyx: actually we should use generics here

amitschendel (approved):

PR #147, opened 2024-09-11
Feature/generic path analyzer

afek854:


matthyx:
can you rebase? this is not normal to have so many commits here - something is wrong

PR #150, opened 2024-09-16
added extra field and tests

afek854:


matthyx:
Let's test if that's they want before merging

PR #152, opened 2024-09-19
add rate limiting per X-Forwarded-For ip

matthyx:


slashben (approved):
It looks good, but it is not the end of the road.

I think we still need a mechanism that limits the maximum number of parallel requests, regardless to the source IP

matthyx:
> It looks good, but it is not the end of the road.
> 
> I think we still need a mechanism that limits the maximum number of parallel requests, regardless to the source IP

I will also make it configurable, and maybe default to no rate limiting

PR #181, opened 2024-12-15
improve searching in CIDR ranges using cidranger

amirmalka:


matthyx (approved):

pkg/apis/softwarecomposition/networkpolicy/v1/networkpolicy.go
```diff
@@ -288,41 +286,32 @@ func generateEgressRule(neighbor softwarecomposition.NetworkNeighbor, KnownServe
...
 	if neighbor.IPAddress != "" {
-		isKnownServer := false
-		// look if this IP is part of any known server
-		for _, knownServer := range KnownServer {
-			for _, entry := range knownServer.Spec {
-				_, subNet, err := net.ParseCIDR(entry.IPBlock)
-				if err != nil {
-					logger.L().Error("error parsing cidr", helpers.Error(err))
-					continue
+		if entries, contains := knownServers.Contains(net.ParseIP(neighbor.IPAddress)); contains {
```
matthyx, line 289: I don't think we need to change v1 since we deprecated it... but since you did already

matthyx (approved):

PR #182, opened 2024-12-24
cleanup application profiles with missing wlid or instance id annotation (when relevancy is enabled)

amirmalka:


matthyx (approved):
interesting concept the multiple handlers, thanks Amir!

PR #194, opened 2025-02-03
pass opts to GetList to allow fullSpec on virtual CRDs

matthyx:
@hebestreit this is what we discussed in slack today, makes sense?


hebestreit:
Yes LGTM and thank you for the fast implementation! 🎉

PR #198, opened 2025-02-24
Removed NetworkNeighbors

jayantpranjal0:
Deprecate NetworkNeighbors


matthyx:
btw, when you create commits, can you use the `-s` option to add the `Signed-off-by:` footer to message?

jayantpranjal0:
Yeahh, I will make sure do it. I actually wanted to resolve the codegen issue first, and then merge the two commits and then signoff.

matthyx:
@jayantpranjal0 why did you remove networkpolicy/v1 and moved up the content of v2 ? This is the surest way of breaking API compabitility.

jayantpranjal0:
I found the difference between networkpolicy/v1 and networkpolicy/v2 to be that v1 implements the functions using NetworkNeighbors, whereas v2 does so through NetworkNeighborhood. There was one more minor change in a function mergeEgressRulesByPorts but both did the same thing, only a minor change in the way an object was created. So, ig v1 was for NetworkNeighbors compatibility. So, I removed it and moved up the contents of v2 to rather just use it as default, as we had to remove NetworkNeighbors completelely.

matthyx:
> I found the difference between networkpolicy/v1 and networkpolicy/v2 to be that v1 implements the functions using NetworkNeighbors, whereas v2 does so through NetworkNeighborhood. There was one more minor change in a function mergeEgressRulesByPorts but both did the same thing, only a minor change in the way an object was created. So, ig v1 was for NetworkNeighbors compatibility. So, I removed it and moved up the contents of v2 to rather just use it as default, as we had to remove NetworkNeighbors completelely.

please don't do that... you can modify the v1 functions to panic but keep the existing paths

jayantpranjal0:
Okayy, I'll modify the v1 functions to panic instead of removing them completely.

PR #224, opened 2025-05-25
Implement queue management system with configurable parameters

slashben:
This pull request introduces a queue management system for handling API requests with configurable queue lengths, worker counts, and maximum object sizes. It includes significant changes across multiple files to implement the new `QueueManager` and integrate it into the server configuration. The key changes are grouped below by theme:

### Queue Management Implementation:
* Added a new `QueueManager` in `pkg/queuemanager/manager.go` to manage per-kind request queues with configurable parameters like queue length, worker count, and maximum object size. It includes logic for request handling, queue creation, and throttling when queues are full.

### Configuration Updates:
* Updated `pkg/config/config.go` to include new configuration fields (`KindQueues`, `DefaultQueueLength`, `DefaultWorkerCount`, `DefaultMaxObjectSize`) and their corresponding default values. Added a new `KindQueueConfig` struct to represent per-kind configurations. [[1]](diffhunk://#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R9-R14) [[2]](diffhunk://#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R28-R33) [[3]](diffhunk://#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R48-R50)
* Added a new `queue-configuration.json` file to define default and per-kind queue configurations.

### Integration with Server:
* Modified `pkg/cmd/server/start.go` to integrate the `QueueManager` into the server's request handling chain by setting `BuildHandlerChainFunc` in the server configuration.
* Added the `queuemanager` package to the imports in `pkg/cmd/server/start.go`.

### Dependency Addition:
* Added the `github.com/panjf2000/ants/v2` library to `go.mod` for managing worker pools in the `QueueManager`.

### DIFF
--- a/build/Dockerfile.debug
+++ b/build/Dockerfile.debug
@@ -1,0 +1,20 @@
+FROM --platform=$BUILDPLATFORM golang:1.23-bookworm AS builder
+
+ENV GO111MODULE=on CGO_ENABLED=0
+WORKDIR /work
+ARG TARGETOS TARGETARCH
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+RUN --mount=target=. \
+    --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg \
+    GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /out/storage .
+
+FROM gcr.io/distroless/static-debian12:debug
+
+COPY --from=builder /go/bin/dlv /usr/bin/dlv
+COPY --from=builder /out/storage /usr/bin/storage
+
+ARG image_version
+ENV RELEASE=$image_version
+
+ENTRYPOINT ["/usr/bin/dlv", "--listen=:40000", "--headless=true","--accept-multiclient", "--continue", "--api-version=2", "--log", "exec", "/usr/bin/storage"]
--- a/pkg/cmd/server/start.go
+++ b/pkg/cmd/server/start.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"time"
 
 	"github.com/didip/tollbooth/v7"
 	"github.com/kubescape/go-logger"
@@ -34,7 +35,9 @@ import (
 	clientset "github.com/kubescape/storage/pkg/generated/clientset/versioned"
 	informers "github.com/kubescape/storage/pkg/generated/informers/externalversions"
 	sampleopenapi "github.com/kubescape/storage/pkg/generated/openapi"
+	"github.com/kubescape/storage/pkg/queuemanager"
 	"github.com/kubescape/storage/pkg/registry/file"
+	"github.com/kubescape/storage/pkg/statscollector"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -54,6 +57,8 @@ import (
 const (
 	defaultEtcdPathPrefix = "/registry/spdx.softwarecomposition.kubescape.io"
 )
+
+var stats = statscollector.NewStatsCollector()
 
 // WardleServerOptions contains state for master/api server
 type WardleServerOptions struct {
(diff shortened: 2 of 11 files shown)


pkg/cmd/server/start.go
```diff
@@ -253,6 +254,13 @@ func (o *WardleServerOptions) Config() (*apiserver.Config, error) {
 	serverConfig.FeatureGate = featuregate.DefaultComponentGlobalsRegistry.FeatureGateFor(featuregate.DefaultKubeComponent)
 	serverConfig.EffectiveVersion = featuregate.DefaultComponentGlobalsRegistry.EffectiveVersionFor(apiserver.WardleComponentName)
 
+	serverConfig.BuildHandlerChainFunc = func(apiHandler http.Handler, c *genericapiserver.Config) http.Handler {
```
matthyx, line 257: nice find, as you see below (line 301) we used `server.GenericAPIServer.Handler.FullHandlerChain` to alter the chain - yours might be a better solution

slashben: I did not want to touch, I wasn't sure what is better

matthyx: definitely, I think it's good this way

pkg/cmd/server/start.go
```diff
@@ -299,6 +317,21 @@ func (o WardleServerOptions) RunWardleServer(ctx context.Context) error {
 		return nil
 	})
 
+	go func() {
+		for {
+			time.Sleep(5 * time.Minute)
+			currentStats := stats.GetStats(true)
+			for kind, verbs := range currentStats {
+				for verb, stats := range verbs {
+					logger.L().Info("stats", helpers.String("kind", kind),
```
matthyx, line 326: do we really want to print this as info?

pkg/queuemanager/manager.go
```diff
@@ -0,0 +1,218 @@
...
+			queue:         queue,
+			pool:          pool,
+			maxObjectSize: maxObjectSize,
+		}
+		logger.L().Info("queue configuration", helpers.String("kind", kind), helpers.Int("queueLength", queueLen), helpers.Int("workerCount", workerCount), helpers.Int("maxObjectSize", maxObjectSize))
+		// Start a dispatcher goroutine for this kind
+		go func(q *kindQueue) {
+			for req := range q.queue {
+				reqLocal := req // capture loop variable
+				err := q.pool.Submit(func() {
```
matthyx, line 95: do you know that this type of queue is blocking if the number of workers is exceeded?

pkg/queuemanager/manager.go
```diff
@@ -0,0 +1,218 @@
...
+					default:
+						// Still active, process
+						reqLocal.next.ServeHTTP(reqLocal.w, reqLocal.r)
+						close(reqLocal.done)
+						return
+					}
+				})
+				if err != nil {
+					logger.L().Error("failed to submit to worker pool", helpers.Error(err), helpers.String("path", reqLocal.r.URL.Path), helpers.String("kind", kind), helpers.String("verb", reqLocal.r.Method), helpers.String("kind", kind))
+					// Optionally: handle the error, e.g., close(req.done) to unblock the waiting goroutine
```
matthyx, line 110: lol

pkg/queuemanager/manager.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+		queue := make(chan *queuedRequest, queueLen)
+		pool, _ := ants.NewPool(workerCount)
+		q = &kindQueue{
+			queue:         queue,
+			pool:          pool,
+			maxObjectSize: maxObjectSize,
+		}
+		logger.L().Info("queue configuration", helpers.String("kind", kind), helpers.Int("queueLength", queueLen), helpers.Int("workerCount", workerCount), helpers.Int("maxObjectSize", maxObjectSize))
```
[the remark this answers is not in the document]
matthyx: well maybe yes actually

pkg/queuemanager/manager.go
```diff
@@ -0,0 +1,218 @@
...
+	qm.mu.Unlock()
+	if !ok || time.Since(lastLog) > 1*time.Minute {
+		logger.L().Warning("queue full for resource", helpers.String("kind", kind), helpers.String("verb", verb), helpers.String("queue", "default"))
+		qm.mu.Lock()
+		qm.lastQueueFullLog[kind] = time.Now()
+		qm.mu.Unlock()
+	}
+}
+
+// TimeoutLoggerMiddleware logs a warning if a request takes longer than 60 seconds to finish.
```
matthyx, line 195: longer than timeoutSeconds

build/Dockerfile.debug
```diff
@@ -0,0 +1,20 @@
+FROM --platform=$BUILDPLATFORM golang:1.23-bookworm AS builder
```
matthyx, line 1: nice!

pkg/queuemanager/manager.go
slashben: In this implementation, the dispatcher goroutine that reads from the queue and submits work to the worker pool can indeed be blocked if the queue is full and all workers are busy. However, in the HTTP handler (QueueHandler), we use a select statement with a default case when attempting to enqueue a request:

```
select {
case q.queue <- req:
    <-req.done
default:
    // handle queue full (e.g., return error to client)
}
```

This means that if the queue is full, the handler will not block; instead, the default case is executed. 

This allows us to fail fast and return an appropriate error response to the client (such as HTTP 429 Too Many Requests), rather than blocking the HTTP request and potentially exhausting server resources.

While the dispatcher can be blocked if the queue is full, the HTTP request handling path is protected from blocking by using the non-blocking select pattern. This ensures that the server remains responsive under load and can handle situations where the queue is at full.

pkg/statscollector/collector.go
```diff
@@ -0,0 +1,110 @@
...
+func extractKindAndVerb(r *http.Request) (kind, verb string) {
+	reqInfo, ok := request.RequestInfoFrom(r.Context())
+	if ok {
+		return reqInfo.Resource, reqInfo.Verb
+	}
+	// fallback:
+	return extractKindAndVerbFromPath(r)
+}
+
+func extractKindAndVerbFromPath(r *http.Request) (kind, verb string) {
```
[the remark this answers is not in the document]
slashben: yes

slashben:
@matthyx please re-review

matthyx (approved):

PR #316, opened 2026-04-23
CollapsConfigs and Wildcard Matching for Opens (and Paths in Endpoints) while reproducing default storage behavior if not used

entlein:
This is the beginning of the storage pr to bring the opensWildcards into upstream
Testing how to create clean PRs that also build with the respective node-agent version

What I want to achieve with this PR:
- Have the CollapsConfig defined
- Have the Defaults all set such that they reproduce the previous behavior
- Have the Wildcard Comparison Function in there, even if not used

I m also running the benchmark test to check that this doesnt incur too much perf-overhead

## Summary by CodeRabbit

* **New Features**
  * Per-prefix path-collapse tuning with adjacent dynamic segments compacted into a single wildcard.
  * Order-aware wildcard-port merging for endpoints; SBOM-provided paths remain preserved.

* **Bug Fixes**
  * Cleanup now skips resources marked as user-managed via annotations.

* **Chores**
  * Adjusted Go module dependency versions.

* **Tests**
  * Extensive new and updated unit tests, benchmarks, and profiling for path/endpoint analysis.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -132,7 +132,7 @@ require (
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.1.1 // indirect
-	github.com/opencontainers/runtime-spec v1.3.0 // indirect
+	github.com/opencontainers/runtime-spec v1.2.1 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
 	github.com/petermattis/goid v0.0.0-20241211131331-93ee7e083c43 // indirect
 	github.com/pierrec/lz4/v4 v4.1.22 // indirect
--- a/go.mod
+++ b/go.mod
@@ -50,7 +50,7 @@ require (
 	github.com/acobaugh/osrelease v0.1.0 // indirect
 	github.com/anchore/go-logger v0.0.0-20250318195838-07ae343dd722 // indirect
 	github.com/anchore/packageurl-go v0.1.1-0.20250220190351-d62adb6e1115 // indirect
-	github.com/anchore/stereoscope v0.1.22 // indirect
+	github.com/anchore/stereoscope v0.1.9 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
 	github.com/armosec/gojay v1.2.17 // indirect
 	github.com/armosec/utils-go v0.0.58 // indirect
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ go 1.25.8
 
 require (
 	github.com/SergJa/jsonhash v0.0.0-20210531165746-fc45f346aa74
-	github.com/anchore/syft v1.42.3
+	github.com/anchore/syft v1.32.0
 	github.com/armosec/armoapi-go v0.0.696
 	github.com/armosec/utils-k8s-go v0.0.30
 	github.com/containers/common v0.63.0
--- a/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
+++ b/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
@@ -10,23 +10,51 @@ import (
 	types "github.com/kubescape/storage/pkg/apis/softwarecomposition"
 )
 
+func isWildcardPort(port string) bool {
+	return port == "0"
+}
+
+func rewritePort(endpoint, wildcardPort string) string {
+	if wildcardPort == "" {
+		return endpoint
+	}
+	port, pathPart := splitEndpointPortAndPath(endpoint)
+	if !isWildcardPort(port) {
+		return ":" + wildcardPort + pathPart
+	}
+	return endpoint
+}
+
 func AnalyzeEndpoints(endpoints *[]types.HTTPEndpoint, analyzer *PathAnalyzer) []types.HTTPEndpoint {
 	if len(*endpoints) == 0 {
 		return nil
 	}
 
-	var newEndpoints []*types.HTTPEndpoint
-	for _, endpoint := range *endpoints {
-		_, _ = AnalyzeURL(endpoint.Endpoint, analyzer)
+	// Detect wildcard port in input (port 0 means any port)
+	wildcardPort := ""
+	for _, ep := range *endpoints {
+		port, _ := splitEndpointPortAndPath(ep.Endpoint)
+		if isWildcardPort(port) {
+			wildcardPort = port
+			break
+		}
 	}
 
+	// First pass: build tree, redirecting to wildcard port if needed
 	for _, endpoint := range *endpoints {
-		processedEndpoint, err := ProcessEndpoint(&endpoint, analyzer, newEndpoints)
+		_, _ = AnalyzeURL(rewritePort(endpoint.Endpoint, wildcardPort), analyzer)
+	}
+
+	// Second pass: process endpoints
+	var newEndpoints []*types.HTTPEndpoint
+	for _, endpoint := range *endpoints {
+		ep := endpoint
+		ep.Endpoint = rewritePort(ep.Endpoint, wildcardPort)
+		processedEndpoint, err := ProcessEndpoint(&ep, analyzer, newEndpoints)
 		if processedEndpoint == nil && err == nil || err != nil {
 			continue
-		} else {
-			newEndpoints = append(newEndpoints, processedEndpoint)
 		}
+		newEndpoints = append(newEndpoints, processedEndpoint)
 	}
 
 	newEndpoints = MergeDuplicateEndpoints(newEndpoints)
@@ -88,6 +116,15 @@ func AnalyzeURL(urlString string, analyzer *PathAnalyzer) (string, error) {
 	return ":" + port + path, nil
 }
 
+func splitEndpointPortAndPath(endpoint string) (string, string) {
+	s := strings.TrimPrefix(endpoint, ":")
+	idx := strings.Index(s, "/")
+	if idx == -1 {
+		return s, "/"
+	}
+	return s[:idx], s[idx:]
+}
+
 func MergeDuplicateEndpoints(endpoints []*types.HTTPEndpoint) []*types.HTTPEndpoint {
 	seen := make(map[string]*types.HTTPEndpoint)
 	var newEndpoints []*types.HTTPEndpoint
@@ -97,10 +134,22 @@ func MergeDuplicateEndpoints(endpoints []*types.HTTPEndpoint) []*types.HTTPEndpo
 		if existing, found := seen[key]; found {
 			existing.Methods = MergeStrings(existing.Methods, endpoint.Methods)
 			mergeHeaders(existing, endpoint)
-		} else {
-			seen[key] = endpoint
-			newEndpoints = append(newEndpoints, endpoint)
+			continue
 		}
+
+		// Check if a wildcard port variant already exists (port 0 means any port)
+		port, pathPart := splitEndpointPortAndPath(endpoint.Endpoint)
+		if !isWildcardPort(port) {
+			wildcardKey := fmt.Sprintf(":%s%s|%s", "0", pathPart, endpoint.Direction)
+			if existing, found := seen[wildcardKey]; found {
+				existing.Methods = MergeStrings(existing.Methods, endpoint.Methods)
+				mergeHeaders(existing, endpoint)
+				continue
+			}
+		}
+
+		seen[key] = endpoint
+		newEndpoints = append(newEndpoints, endpoint)
 	}
 
 	return newEndpoints
@@ -111,7 +160,6 @@ func getEndpointKey(endpoint *types.HTTPEndpoint) string {
 }
 
 func mergeHeaders(existing, new *types.HTTPEndpoint) {
-	// TODO: Find a better way to unmashal the headers
 	existingHeaders, err := existing.GetHeaders()
 	if err != nil {
 		return
--- a/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
+++ b/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
@@ -109,18 +109,22 @@ func splitEndpointPortAndPath(endpoint string) (string, string) {
 }
 
 // MergeDuplicateEndpoints folds duplicates and merges same-path specific-port
-// endpoints into a wildcard-port (:0) sibling. The folding is symmetric:
+// endpoints into a wildcard-port (:0) sibling. Folding is symmetric and is
+// keyed on the same triple HTTPEndpoint.Equal compares — (Endpoint,
+// Direction, Internal). An Internal=false endpoint will therefore NOT merge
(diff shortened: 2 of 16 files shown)


entlein:
The first commit had the following baseline test

> go test -bench='^BenchmarkAnalyzePath(|Warm|Cold)$' -benchmem -benchtime=3s ./pkg/registry/file/dynamicpathdetector/tests/
> ?       github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector      [no test files]
> ok      github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests        2.359s
> goos: darwin
> goarch: amd64
> pkg: github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests
> cpu: VirtualApple @ 2.50GHz
> BenchmarkAnalyzePath-14         14158999               247.8 ns/op            39 B/op          1 allocs/op
> BenchmarkAnalyzePathWarm-14     14414613               246.6 ns/op            39 B/op          1 allocs/op
> BenchmarkAnalyzePathCold-14      4350681               831.9 ns/op          1913 B/op         21 allocs/op

entlein:
> go test -count=1 -timeout=2m ./pkg/registry/file/dynamicpathdetector/...
> go test -bench='^BenchmarkAnalyzePathWarm$' -benchmem -benchtime=3s -run=^$ ./pkg/registry/file/dynamicpathdetector/tests/
> ?       github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector      [no test files]
> ok      github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests        2.442s
> goos: darwin
> goarch: amd64
> pkg: github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests
> cpu: VirtualApple @ 2.50GHz
> BenchmarkAnalyzePathWarm-14     13851844               257.3 ns/op            39 B/op          1 allocs/op
> PASS
> ok      github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests        6.065s

entlein:
pkg: github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests
cpu: VirtualApple @ 2.50GHz
BenchmarkAnalyzePathWarm-14     11753203               307.5 ns/op            39 B/op          1 allocs/op

entlein:
pkg: github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests
cpu: VirtualApple @ 2.50GHz
BenchmarkAnalyzePathWarm-14     10592332               334.0 ns/op            39 B/op          1 allocs/op

entlein:
> go test -count=1 -timeout=2m ./pkg/registry/file/dynamicpathdetector/...
> go test -bench='^BenchmarkAnalyzePathWarm$' -benchmem -benchtime=3s -run=^$ ./pkg/registry/file/dynamicpathdetector/tests/
> ?       github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector      [no test files]
> ok      github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests        2.665s
> goos: darwin
> goarch: amd64
> pkg: github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests
> cpu: VirtualApple @ 2.50GHz
> BenchmarkAnalyzePathWarm-14     10105309               339.7 ns/op            39 B/op          1 allocs/op
> PASS
> ok      github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector/tests        6.023s

entlein:
Component tests green here https://github.com/k8sstormcenter/node-agent/actions/runs/24909769622/job/72949695747  . The randomx-test is red for totally unrelated reasons.

go.mod
```diff
@@ -6,7 +6,7 @@ go 1.25.8
 
 require (
 	github.com/SergJa/jsonhash v0.0.0-20210531165746-fc45f346aa74
-	github.com/anchore/syft v1.42.3
+	github.com/anchore/syft v1.32.0
```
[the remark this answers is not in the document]
matthyx: we cannot upgrade syft right now, it'll be part of an epic next month covering storage/node-agent/kubevuln/kubescape and their respective system tests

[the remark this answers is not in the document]
matthyx: no thanks Rabbit

pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
```diff
@@ -10,23 +10,51 @@ import (
...
+		if isWildcardPort(port) {
+			wildcardPort = port
+			break
+		}
+	}
+
+	// First pass: build tree, redirecting to wildcard port if needed
 	for _, endpoint := range *endpoints {
-		_, _ = AnalyzeURL(endpoint.Endpoint, analyzer)
+		_, _ = AnalyzeURL(rewritePort(endpoint.Endpoint, wildcardPort), analyzer)
```
matthyx: _⚠️ Potential issue_

`wildcardPort` is computed once for the whole slice and then applied to every endpoint in both passes. That means a single `:0/health` entry rewrites unrelated concrete ports too: `[:0/health, :443/login]` becomes `[:0/health, :0/login]`, which over-broadens the generated profile.

Suggested fix: don't pre-normalize the entire slice to the first wildcard port. Keep the original port when feeding the analyzer, then only fold a concrete port into `:0` when you have a wildcard sibling for the **same path + direction** (for example in `MergeDuplicateEndpoints` after the analyzed path is known). That preserves the intended “`:0` absorbs matching specifics” behavior without turning unrelated endpoints into any-port matches.

entlein: Thanks for the review, I need to write a suite of tests for this. Thanks for catching this!

entlein: commit https://github.com/k8sstormcenter/storage/commit/54e71796404deb747382147ee7451c9432ca0343 to address the isse

pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
```diff
@@ -97,21 +130,67 @@ func MergeDuplicateEndpoints(endpoints []*types.HTTPEndpoint) []*types.HTTPEndpo
...
+			}
 			seen[key] = endpoint
 			newEndpoints = append(newEndpoints, endpoint)
+			_ = absorbed
+			continue
+		}
+
+		// Specific port: if a wildcard sibling for the same (path, direction)
+		// is already in `seen`, fold this entry into it.
+		wildcardKey := fmt.Sprintf(":0%s|%s", pathPart, endpoint.Direction)
```
matthyx: Potential issue: wildcard folding here matches only on path + direction, but `HTTPEndpoint.Equal` also distinguishes `Internal`. As written, `:0/login` with `Internal=false` can absorb `:443/login` with `Internal=true` and collapse them into one endpoint. Can we include `Internal` in this key and in `getEndpointKey()`?

pkg/registry/file/dynamicpathdetector/tests/analyze_endpoints_test.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func TestAnalyzeEndpoints(t *testing.T) {
-	analyzer := dynamicpathdetector.NewPathAnalyzer(100)
+	analyzer := dynamicpathdetector.NewPathAnalyzerWithConfigs(dynamicpathdetector.EndpointDynamicThreshold, nil)
```
matthyx: Potential issue: this table test reuses one mutable `PathAnalyzer` across all subtests. The new `Test with 0 port` case only passes after earlier cases have already trained the trie; run alone it still fails. Can we create a fresh analyzer inside each `t.Run` (or make the warm-state setup explicit in that case)?

pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
entlein: addressed in https://github.com/kubescape/storage/pull/316/commits/dd028cf30f706c480d95f526d66618c240371438

pkg/registry/file/dynamicpathdetector/analyzer.go
```diff
@@ -134,33 +347,82 @@ func shallowChildrenCopy(src, dst *SegmentNode) {
...
-		for regularIndex < regularLen && regularPath[regularIndex] != '/' {
-			regularIndex++
+func compareSegments(dynamic, regular []string) bool {
+	if len(dynamic) == 0 {
+		return len(regular) == 0
+	}
+	if dynamic[0] == WildcardIdentifier {
+		// A trailing `*` matches any remaining path tail (including empty).
+		if len(dynamic) == 1 {
+			return true
```
matthyx: Behavior regression: `CompareDynamic("/etc/*", "/etc")` used to return `false` and now it's `true`. Was it intended?
For example if we have `/etc/*` R0002 will allow accessing the directory `/etc` itself, where previously it was flagged as anomalous.

pkg/registry/file/dynamicpathdetector/analyzer.go
```diff
@@ -134,33 +347,82 @@ func shallowChildrenCopy(src, dst *SegmentNode) {
...
-	dynamicLen, regularLen := len(dynamicPath), len(regularPath)
-
-	for dynamicIndex < dynamicLen && regularIndex < regularLen {
-		// Find the next segment in dynamicPath
-		dynamicSegmentStart := dynamicIndex
-		for dynamicIndex < dynamicLen && dynamicPath[dynamicIndex] != '/' {
-			dynamicIndex++
-		}
-		dynamicSegment := dynamicPath[dynamicSegmentStart:dynamicIndex]
+	dynamicSegments := strings.Split(dynamicPath, "/")
```
matthyx: Performance regression: the old implementation was a manual index scan — zero heap allocations per call. The new code allocates two `[]string` slices on every invocation. And this is called on every file-open per R0002.
Solution: port the old two-pointer approach to handle `*`.

pkg/registry/file/dynamicpathdetector/types.go
```diff
@@ -1,6 +1,51 @@
...
+// DefaultCollapseConfigs carries the per-prefix thresholds we've found
+// useful in practice. These are the defaults wired into AnalyzeOpens; a
+// caller can pass a different slice via NewPathAnalyzerWithConfigs if
+// they want to tune for their workload.
+var DefaultCollapseConfigs = []CollapseConfig{
+	{Prefix: "/etc", Threshold: 100},
+	{Prefix: "/etc/apache2", Threshold: 50}, // tuned for the webapp standard test
+	{Prefix: "/opt", Threshold: 50},
+	{Prefix: "/var/run", Threshold: 50},
+	{Prefix: "/app", Threshold: 50}, // any variation under /app collapses immediately
```
matthyx:
```suggestion
	{Prefix: "/app", Threshold: 50}, // any variation under /app collapses at 50 unique children
```

pkg/registry/file/dynamicpathdetector/types.go
```diff
@@ -1,6 +1,51 @@
...
+type CollapseConfig struct {
+	Prefix    string
+	Threshold int
+}
+
+// DefaultCollapseConfigs carries the per-prefix thresholds we've found
+// useful in practice. These are the defaults wired into AnalyzeOpens; a
+// caller can pass a different slice via NewPathAnalyzerWithConfigs if
+// they want to tune for their workload.
+var DefaultCollapseConfigs = []CollapseConfig{
```
matthyx:
```suggestion
var defaultCollapseConfigs = []CollapseConfig{
```
and create a func: `DefaultCollapseConfigs() []CollapseConfig { return append([]CollapseConfig(nil), defaultCollapseConfigs...) }`

matthyx (changes requested):
2 regressions and some nits

pkg/registry/file/dynamicpathdetector/analyzer.go
```diff
@@ -134,33 +347,86 @@ func shallowChildrenCopy(src, dst *SegmentNode) {
...
+		return len(regular) == 0
+	}
+	if dynamic[0] == WildcardIdentifier {
+		// A trailing `*` matches one OR MORE remaining segments — never
+		// zero. Standard glob semantics: `/etc/*` matches files under
+		// /etc but NOT the bare /etc directory itself. Without this,
+		// R0002 would silently allow access to a profiled directory's
+		// parent, which masks tampering of the directory entry itself.
+		if len(dynamic) == 1 {
+			return len(regular) > 0
```
matthyx, line 408: Potential issue: the new trailing-`*` restriction still lets the real root path through. `CompareDynamic("/*", "/")` returns `true` because `strings.Split("/", "/")` yields `["", ""]`, so by the time this branch runs `len(regular) > 0`. The new regression test checks `regular: ""` instead of `"/"`, so it misses the actual root-path input. Can we normalize `/` before matching (or trim empty split segments consistently) and add an explicit `CompareDynamic("/*", "/") == false` test?

pkg/registry/file/dynamicpathdetector/analyzer.go
matthyx: @entlein I'd like to fix this one

pkg/registry/file/dynamicpathdetector/analyzer.go
```diff
@@ -134,33 +347,86 @@ func shallowChildrenCopy(src, dst *SegmentNode) {
...
-	for dynamicIndex < dynamicLen && regularIndex < regularLen {
-		// Find the next segment in dynamicPath
-		dynamicSegmentStart := dynamicIndex
-		for dynamicIndex < dynamicLen && dynamicPath[dynamicIndex] != '/' {
-			dynamicIndex++
-		}
-		dynamicSegment := dynamicPath[dynamicSegmentStart:dynamicIndex]
+	dynamicSegments := strings.Split(dynamicPath, "/")
+	regularSegments := strings.Split(regularPath, "/")
+	return compareSegments(dynamicSegments, regularSegments)
```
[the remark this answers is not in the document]
matthyx: @coderabbitai are you sure it's not already addressed?

pkg/registry/file/dynamicpathdetector/types.go
```diff
@@ -1,6 +1,59 @@
...
+func DefaultCollapseConfigs() []CollapseConfig {
+	return append([]CollapseConfig(nil), defaultCollapseConfigs...)
+}
+
+// DefaultCollapseConfig is the global fallback used when no CollapseConfig
+// prefix matches the walked path.
+var DefaultCollapseConfig = CollapseConfig{
+	Prefix:    "/",
+	Threshold: OpenDynamicThreshold,
+}
```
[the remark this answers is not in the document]
matthyx: @coderabbitai isn't this one stale?

pkg/registry/file/dynamicpathdetector/analyzer.go
entlein: Yes, I like the explicit requirement, that `CompareDynamic("/*", "/") == false` and means that a profile needs to explicitely allowlist `CompareDynamic("*", "/") == true` if they wish to allow the root path itself.

Issue #315, opened 2026-04-22
Aggregate user-managed (ug- prefix) profiles into the consolidated ContainerProfile


matthyx:
## Context

The node-agent's unified `ContainerProfileCache` (see [kubescape/node-agent#788](https://github.com/kubescape/node-agent/pull/788)) is CP-direct: it reads the consolidated `ContainerProfile` at the stable `GetSlug(false)` name as the single source of truth. The storage server already aggregates the per-tick time-series `ContainerProfile` writes (named `<slug>-<uuid>`) into a consolidated CP at the `<slug>` name — that part works well.

What it does NOT aggregate is the **user-managed** profiles published at the `"ug-" + workloadSlug` well-known name. These are authored by humans (either directly or via the kubescape UI) and carry the annotation:

```yaml
kubescape.io/managed-by: User
```

The legacy `ApplicationProfileCache.handleUserManagedProfile` auto-discovered these via the annotation + `ug-` prefix and merged them into the workload's cached profile (see [legacy source, deleted in the migration PR](https://github.com/kubescape/node-agent/blob/900e72bd/pkg/objectcache/applicationprofilecache/applicationprofilecache.go#L269-L351)).

## Current state in the new cache

As a short-term fix in PR #788, the node-agent fetches `ug-<workloadName>` AP and NN on every `addContainer` and on every reconciler refresh tick, then projects them on top of the consolidated CP. This works but:

- Adds 2 RPCs per container per refresh tick (on top of the 1 CP GET).
- Duplicates the merge logic between node-agent and (eventually) storage.
- Splits the \"CP-direct\" principle: the CP isn't actually authoritative because the cache still has to go fetch user-managed profiles separately.
- Every agent on every node does the merge independently; server-side aggregation would do it once.

## Proposal

When the storage server consolidates time-series `ContainerProfile` writes into the stable-named CP (via `ContainerProfileProcessor.consolidateKeyTimeSeries`), **also merge the matching user-managed profile** (`ug-<workloadSlug>` if it exists) into the output.

Matching rules should mirror what `handleUserManagedProfile` did:

- Workload slug: `instanceID.GetSlug(true)` (no container name).
- Per-container: find the matching `ApplicationProfileContainer` / `NetworkNeighborhoodContainer` by `Name` across `.Spec.Containers + .Spec.InitContainers + .Spec.EphemeralContainers` and merge its fields into the consolidated CP's flat spec.
- Completion / Status annotations come from the base CP; user-managed data is additive.

[...]

PR #319, opened 2026-05-07
feat(processor): merge ug- AP/NN into consolidated ContainerProfile

yugal07:
Server-side aggregation of user-managed (ug-prefix) ApplicationProfile and                                       
  NetworkNeighborhood into the consolidated ContainerProfile during                                                
  consolidateKeyTimeSeries. Lets node-agent treat the CP as a single source                                        
  of truth and drop its per-tick ug- fetch.                                                                        
                                                                                                                   
  - Merge math ported from node-agent's projection.go, operating on the                                            
    internal softwarecomposition types (no interface or storage method                                             
  - Idempotent across ticks via kubescape.io/last-merged-ug-{ap,nn}-rv                                             
    annotation markers; merge skipped when ug- ResourceVersion is unchanged.                                       
  - Re-consolidation trigger: option (c) (#315 discussion),  relies on the existing 30s                                              
    ConsolidateTimeSeries cadence; no AP/NN watcher added (per @matthyx                                     
    guidance, memory cost of caching ug- objects rules out a watch path).                                          
  - Deterministic ordering for DNSNames and MatchExpressions to keep                                               
    SyncChecksum stable across re-merges of unchanged content.                                                     
  - LabelSelector / PodSelector / NamespaceSelector MatchLabels: user value                                        
    wins on collision (utils.MergeMaps preserves base, so a local                                                  
    overrideMerge is used here instead).                                                                           
                                                                                                                   
  Refs kubescape/storage#315, kubescape/node-agent#788                                                             
                                                                                                                   
  Known limitations (to discuss in PR review):              
  - Stale-on-edit/delete: merge is additive against the persisted CP, so                                           
    removing entries from a ug- CRD does not retract them from already- 
    merged consolidated CPs. The proper fix is a CRD shape change                                                  
    (separate user-managed sub-spec, replaced wholesale)                                         
  - Idle workloads: updateProfile early-returns when no new time-series                                            
    arrives, so ug- edits to Completed workloads do not propagate until                                            
    the next tick that has new data. Accepted trade-off for option (c).                                            
                                 
  worth a note: Changes were initially drafted using help from copilot, and manually reviewed and edited

## Summary by CodeRabbit

* **New Features**
  * Produces per-container "merged" profiles by combining observed data with user-managed application and network overlays; registers merged-first REST reads and a separate merged storage variant.

* **Bug Fixes**
  * Avoids unnecessary observed writes when no new time-series data; skips processing when required identifiers are missing; retracts stale merged profiles when overlays are removed; refreshes merged state each tick without forcing observed saves.

* **Tests**
  * Added extensive unit and e2e tests covering merge correctness, idempotency, RV behavior, merged-first semantics, retraction, cleanup, and edge cases.


yugal07:
@matthyx. Let me know what all things needs to change. :smile:

matthyx:
> @matthyx. Let me know what all things needs to change. 😄

I'll try to review it tomorrow, it's a lot of mental load

yugal07:
> > @matthyx. Let me know what all things needs to change. 😄
> 
> I'll try to review it tomorrow, it's a lot of mental load

No issues. Thanks

yugal07:
@matthyx Is this PR still relevant? I'd be happy to make the changes discussing and fixing the limitations mentioned / other changes..

matthyx:
I think the safer direction here is to **stop mutating the canonical observed CP in place** and instead make the merged view an explicit derived artifact.

Concretely, I’d suggest implementing it in these steps:

1. Keep the current `containerprofile` key as the **observed/canonical** profile produced only from time-series consolidation.
2. Build a separate **merged/effective CP** from `observed CP + current ug- AP/NN overlay`, and store it under a different key/location.
3. Make the split explicit in the storage/read API so the consolidator path (`loadOrInitializeProfile` / `updateProfile`) always reads and writes the **observed** CP, while consumers read the effective one. I would avoid relying on a single `GetContainerProfile()` path to serve both roles by convention.
4. Change the normal consumer load path to **prefer the merged CP and fall back to the observed CP** when the merged artifact does not exist.
5. Refresh or delete the merged artifact even when there is **no new TS data**, so ug edits/deletes for idle or `Completed` workloads still propagate correctly. In other words, the current `!newData` early return cannot fully short-circuit the derived merged write path.
6. Keep aggregated AP/NN derived from the **effective/merged** CP so all downstream outputs stay aligned with what node-agent actually reads.
7. Add explicit metadata on the merged CP so it is easy to identify/debug:
   - a simple label marking it as a merged profile
   - annotation(s) pointing back to the relevant `ug-` source profile(s) / key(s), and ideally the source RV(s)

   My bias would be: use the label only for quick identification/filtering, and keep the richer source reference in annotations.
8. Add tests for:
   - ug edit/delete retracting data from the merged CP
   - idle/no-new-data refresh/delete behavior
   - merged-read fallback to observed when merged is absent
   - consolidator never re-reading the merged artifact

That should give us the provenance we need and fix the stale-on-edit/delete semantics without continuing to mutate the canonical observed CP.

yugal07:
Hi @matthyx, thanks a lot for the detailed review. I pushed a follow-up commit that reworks the design along the lines you suggested.

Quick summary of what changed:

* The observed CP is now only written from time-series consolidation. The `ug-` merge never touches it.
* The merged (effective) CP lives under a parallel key (`containerprofile-merged`), built by a new `MergedKeyFor` helper.
* Added `Save/Get/DeleteMergedContainerProfile` to the storage interface so the split is explicit rather than convention-based.
* Wrapped the apiserver read path with `ContainerProfileRESTStorage`: `GET` tries merged first and falls back to observed on not-found. `List/Watch/Update` pass through unchanged. `Delete` cascades to the merged sibling.
* The merged artifact is rebuilt from scratch every tick from `(observed, ug-AP, ug-NN)`, so edits and deletes retract naturally without RV markers. When no `ug-` overlay exists, the merged key is deleted so consumers fall back to observed.
* The merged refresh now runs even when there is no new TS data for the tick, so `ug-` edits propagate to idle workloads as long as the consolidator still visits them.
* Aggregated AP/NN now derive from the effective CP so downstream consumers all see the same view.
* Added provenance metadata on merged CPs:

  * label: `kubescape.io/profile-kind=merged`
  * annotations carrying the source `ug-` keys, their `ResourceVersions`, the observed RV, and merge timestamp.

Tests added cover:

* retraction on `ug-` delete
* refresh on no-new-data
* merged-first REST fallback
* consolidator never reading the merged key (poison test)
* full E2E walkthrough across the `ug-` lifecycle

All existing tests still pass.

Two things I’d like your call on before considering this ready:

1. **Orphan cleanup**
   The `cleanup()` map in `containerprofile_processor.go` does not currently include the new `containerprofile-merged` kind. The REST `Delete` path and the per-tick refresh handle the common cases, but a workload that gets age-cleaned without going through `Delete` could leave a merged orphan.

   Would you prefer I add it to `resourceToKindHandler`, or leave cleanup to the refresh path?

2. **NN selector behavior on container-name mismatch**
   Right now, if the `ug-` NN has no matching container, the pod `LabelSelector` still merges at the workload level. I kept the legacy behavior so a typo in the `ug-` still propagates the selector, but I can switch it to strict per-container semantics if that’s preferable.

matthyx:
1. add it to `resourceToKindHandler`
2. keep the ug- at workload level, since it's an application profile, it contains a list of all init/ephemeral/regular containers, I don't see the issue here

Also make merged-key delete failures hard errors / retryable instead of swallowing them.

pkg/registry/file/containerprofile_user_managed_test.go
```diff
@@ -0,0 +1,1195 @@
...
+	}
+	h.seedNonCP(e2eUgAPKey(), userAP)
+
+	h.consolidate()
+	first := h.requireMerged()
+	require.Equal(t, 1, count(first.Spec.Capabilities, "USER_MANAGED_CAP"))
+	rvAfterFirst := first.Annotations[mergedSourceUserAPRVKey]
+	require.NotEmpty(t, rvAfterFirst)
+
+	// Second tick: introduce a new TS profile (p2) so updateProfile re-runs
```
matthyx: you should:

- do a second tick without adding data and make sure the "identical" CP isn't written
- add the new data for a third tick

then fix the code to make the test pass :-)

pkg/registry/file/containerprofile_storage.go
```diff
@@ -113,6 +135,40 @@ func (c *ContainerProfileStorageImpl) SaveContainerProfile(ctx context.Context,
...
+func (c *ContainerProfileStorageImpl) GetMergedContainerProfile(ctx context.Context, observedKey string) (softwarecomposition.ContainerProfile, error) {
+	conn := ctx.Value(connKey).(*sqlite.Conn)
+	profile := softwarecomposition.ContainerProfile{}
+	err := c.storageImpl.GetWithConn(ctx, conn, MergedKeyFor(observedKey), storage.GetOptions{}, &profile)
+	return profile, err
+}
+
+func (c *ContainerProfileStorageImpl) DeleteMergedContainerProfile(ctx context.Context, observedKey string) error {
+	conn := ctx.Value(connKey).(*sqlite.Conn)
+	err := c.storageImpl.delete(ctx, conn, MergedKeyFor(observedKey), &softwarecomposition.ContainerProfile{}, nil, nil, nil, storage.DeleteOptions{})
```
matthyx: you should probably use `DeleteWithConn` to be consistent with Get and Save
(and avoid an un-synchronized delete could lead to database lock contention or SQLite busy errors)

pkg/registry/file/containerprofile_user_managed_test.go
matthyx: hint for the fix `merged-at`

matthyx: I think it might also fail because of `observed.ResourceVersion`

pkg/registry/file/containerprofile_processor.go
```diff
@@ -464,34 +469,101 @@ func (a *ContainerProfileProcessor) updateProfile(ctx context.Context, timeSerie
...
+	if err != nil {
+		return observed, err
+	}
+
+	if !hasOverlay {
+		// No ug- input. Delete any prior merged artifact so consumers fall
+		// back to observed. Probe first so the no-merged-yet path (every
+		// workload's first tick before any ug- exists) doesn't issue a futile
+		// delete that the storage layer logs at Error level.
+		if _, getErr := a.ContainerProfileStorage.GetMergedContainerProfile(ctx, observedKey); getErr != nil {
```
matthyx: see if you can avoid the extra lock here, just to avoid the error log

yugal07:
On it. Thanks

yugal07:
@matthyx 
Idempotency - The merged CP was a DeepCopy of observed, so it carried observed's ResourceVersion and SyncChecksum, and SaveMergedContainerProfile passed an empty cachedExistingObject so GuaranteedUpdate never compared against the stored object. Switched to nil and carry the persisted merged object's identity forward, so an unchanged rebuild is a no-op. Also dropped the per-tick merged-at annotation.

DeleteMergedContainerProfile now uses DeleteWithConn for the synchronized delete, fronted by a lock-free ReadMetadata check so the no-ug case stays quiet without the extra lock, and refreshMergedProfile drops its own probe.

Test is now three ticks: create, a no-data tick asserting the merged CP is not rewritten (zero watch events, stable RV, byte identical), then a change tick asserting it is.

Two things for your read. The third tick edits the ug-AP rather than feeding new TS data, because the no-data tick relies on the expired path which marks the workload Completed, and that then rejects new TS profiles. Observed-change is already covered by the E2E and RV-bump tests, but I can reorder to create, new TS data, then no-data if you prefer that. And on scope: this fixes the idle-tick churn, but the merged still refreshes once per node-agent report since observed re-saves and bumps RV on every report. Insulating that means making the observed save idempotent too. What do you think? Should i push a follow up commit for this?

matthyx:
> but the merged still refreshes once per node-agent report since observed re-saves and bumps RV on every report

yes please fix that too

yugal07:
@coderabbitai review

matthyx (approved):
ready to release, we'll have to make sure this one is available to all until we remove the merge from node-agent: https://github.com/kubescape/node-agent/issues/830

PR #324, opened 2026-05-15
allowing CIDRs, wildcards and Plural in IP and DNS

entlein:
Sorry, we do not accept changes directly against this repository. Please see
CONTRIBUTING.md for information on where and how to contribute instead.

## Summary by CodeRabbit

* **New Features**
  * Added `NetworkNeighbor.IPAddresses` (list) support for literal IPs, CIDRs, and `*`, including wildcard-aware matching.
  * Added wildcard-aware DNS name matching (including `⋯` and `*` rules).
* **Validation & API Changes**
  * Added admission-time validation for network profile IP/DNS entries, including deprecated single-value fields when set.
* **Documentation**
  * Documented the IP/DNS matching contract and performance guidance.
* **Tests**
  * Added protobuf encoding/round-trip tests plus unit tests and benchmarks for matching and validation.
* **Chores**
  * Bumped the protoc Docker build image version.
* **Deprecated**
  * Deprecated singular `IPAddress`; use `IPAddresses` instead.

PR #325, opened 2026-05-15
profile-compaction: CollapseConfig CRD + projection overlay + user-maged lifecycle

entlein:
- `CollapseConfig` CRD (v1beta1 with conversion, deepcopy, protobuf, openapi, generated clientset/informers/listers, etcd registration)
    lets cluster operators declare path-prefix policies that drive profile-compaction without code changes
  - `applicationprofile_processor` + `containerprofile_processor` consume the CollapseConfig provider to choose per-profile collapse strategies at admission time
  - `pkg/registry/file/cleanup.go`: cleanup loop skips user-managed resources (annotation-based marker, not labels —
    labels would silently miss every user-managed profile)
  - `artifacts/collapseconfiguration-default-sample.yaml` ships a sample CR

## Summary by CodeRabbit

* **New Features**
  * Added a cluster-scoped `CollapseConfiguration` CRD (including a default sample) to control dynamic path-collapse thresholds, with per-prefix longest-prefix-wins overrides.
  * Dynamic collapse settings are now applied during profile deflation.
  * Network neighbor profiles now support `IPAddresses` (list form) for wildcard-aware IP matching, alongside expanded NetworkNeighborhood admission validation.

* **Bug Fixes**
  * Resource cleanup now preserves user-managed resources.
  * Watch behavior now prefers “idle” watches to avoid tight re-watch loops.

* **Documentation**
  * Added IP/DNS wildcard matching docs and updated README/watchlist/requirements guidance.

* **Tests**
  * Expanded coverage for CRD translation, protobuf compatibility, admission validation, and network matching correctness.


entlein:
@coderabbitai review

Collective re-review request after applying your atomic-split advisory + rabbit-feedback fixes across the full SBOB upstream PR set. Each PR is now scoped to a single tier per your dependency map:

Tier 0 (opens):
- kubescape/storage#323 (path-wildcards core: anchored trailing-*, per-port key, WildcardIdentifier + CollapseConfig API — squash per your "merge first / or squash together" note)
- kubescape/node-agent#811 (was_path_opened* CEL helpers — NEW)

Tier 2 (execs):
- kubescape/storage#322 (CompareExecArgs)
- kubescape/storage#325 (CollapseConfig CRD + processors)
- kubescape/node-agent#805 (resolveExecPath + 3-arg parse)
- kubescape/node-agent#807 (R0040 args consumer)
- kubescape/node-agent#812 (ExecsByPath composite-key surface — NEW)

Tier 3 (network):
- kubescape/storage#324 (networkmatch package)
- kubescape/node-agent#806 (network wildcards CEL)

Tier 4 (tamper/sign):
- kubescape/node-agent#808 (projection overlay + tamper + non-blocking fanout + NN resource)
- kubescape/node-agent#809 (cosign sign/verify + tamper detection)

Umbrellas:
- kubescape/storage#326
- kubescape/node-agent#810

Please confirm atomicity per your earlier advisory and flag any remaining cross-PR coupling we missed.

pkg/apis/softwarecomposition/zz_generated.deepcopy.go
```diff
@@ -1970,6 +2067,11 @@ func (in *NetworkNeighbor) DeepCopyInto(out *NetworkNeighbor) {
 		*out = new(metav1.LabelSelector)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.IPAddresses != nil {
```
matthyx, line 2070: Blocking: this generated deepcopy now expects `NetworkNeighbor.IPAddresses`, but both internal and v1beta1 `NetworkNeighbor` still expose `IPAddress` (singular). `go test ./pkg/apis/softwarecomposition` fails immediately here. Unless the field rename is included in this PR/rebased prerequisite, this codegen drift leaves the branch unbuildable.

pkg/registry/file/applicationprofile_processor.go
```diff
@@ -108,13 +137,13 @@ func (a *ApplicationProfileProcessor) SetStorage(containerProfileStorage Contain
 	a.storageImpl = containerProfileStorage
 }
 
-func deflateApplicationProfileContainer(container softwarecomposition.ApplicationProfileContainer, sbomSet mapset.Set[string]) softwarecomposition.ApplicationProfileContainer {
-	opens, err := dynamicpathdetector.AnalyzeOpens(container.Opens, dynamicpathdetector.NewPathAnalyzer(OpenDynamicThreshold), sbomSet)
+func deflateApplicationProfileContainer(container softwarecomposition.ApplicationProfileContainer, sbomSet mapset.Set[string], settings dynamicpathdetector.CollapseSettings) softwarecomposition.ApplicationProfileContainer {
+	opens, err := dynamicpathdetector.AnalyzeOpens(container.Opens, dynamicpathdetector.NewPathAnalyzerWithConfigs(settings.OpenDynamicThreshold, settings.CollapseConfigs), sbomSet)
```
matthyx, line 141: Blocking: this PR now calls `dynamicpathdetector.NewPathAnalyzerWithConfigs(...)`, but that symbol (along with `CollapseConfig`, `DefaultCollapseConfigs`, and the threshold constants used by `collapse_config_from_crd.go`) does not exist on current `main`. As-is, the branch still depends on an unmerged prerequisite and cannot compile until those path-wildcards changes are rebased/cherry-picked in.

pkg/apiserver/apiserver.go
```diff
@@ -160,6 +161,7 @@ func (c completedConfig) New() (*WardleServer, error) {
 	)
 	apiGroupInfo.VersionedResourcesStorageMap["v1beta1"] = map[string]rest.Storage{
 		"applicationprofiles":                 ep(applicationprofile.NewREST, applicationProfileStorageImpl),
+		"collapseconfigurations":              ep(collapseconfiguration.NewREST),
```
matthyx, line 186: Blocking: registering the CRD endpoint here does not wire it into the live deflate path. `applicationProfileStorageImpl` and `containerProfileStorageImpl` are still constructed with the default collapse-settings providers, and I cannot find any non-test call to `SetCollapseSettings(...)` / assignment to `CollapseSettings`. So `CollapseConfiguration/default` is stored, but never consulted when profiles are compacted; applying the sample manifest is a no-op.

matthyx (changes requested):
I found three merge blockers:

- the generated deepcopy drift now references `NetworkNeighbor.IPAddresses` while the handwritten API types still expose `IPAddress`, so the branch does not build;
- this branch calls dynamicpathdetector APIs that are not on current `main` (`NewPathAnalyzerWithConfigs`, `CollapseConfig`, `DefaultCollapseConfigs`, etc.), so it still depends on an unmerged prerequisite;
- the new `CollapseConfiguration` resource is registered, but the application/container profile processors still use compiled-in defaults, so the CRD is never read in production and the sample manifest is a no-op.

Please fix those before merge.

matthyx (commented):
Rechecked after the latest push: the three blockers from my previous review are fixed, but I found two new blockers.

- `pkg/apis/softwarecomposition/networkpolicy/v2/networkpolicy.go:340` and `:419` still generate peers/policy refs only from the deprecated singular `neighbor.IPAddress` / `neighbor.DNS` fields. The new `ipAddresses` payload added in this PR is never consumed here, so a profile that uses only `ipAddresses` produces no corresponding `IPBlock` entries in the generated NetworkPolicy.
- `pkg/registry/file/networkneighborhood_processor.go:91` merges duplicate neighbors by `Identifier`, but on merge it only appends `DNSNames` and `Ports`. The new `IPAddresses` slice is not appended/deduped, so when multiple observations for the same identifier contribute different IPs, every IP except the first neighbor's slice is dropped before the profile is stored.

matthyx:
One more blocker after checking the generated artifacts: the API type now exposes `ipAddresses`, but the generated surfaces are still stale.

- `pkg/generated/applyconfiguration/softwarecomposition/v1beta1/networkneighbor.go` still only has `IPAddress` / `WithIPAddress`, so server-side-apply builders cannot express the new field.
- `pkg/generated/openapi/zz_generated.openapi.go` still serves only `ipAddress` and even keeps it in the required list, so the published schema is out of sync with the updated Go type.

Please regenerate the applyconfiguration/OpenAPI artifacts for `NetworkNeighbor` as part of this change.

entlein:
@coderabbit, please review all PRs 328 324 and 325 together now

entlein:
Pushed a fix for the request-path stall flagged in review (commit `a6fce873`).

**Root cause:** the `CollapseConfig` provider performed an unbounded `s.Get(context.Background(), …)` per deflate call against the apiserver's own storage. Under component-test load (every system pod produces a `ContainerProfile`) this self-referential read stalled the request path, so application profiles were never produced and profile-dependent tests timed out (post-timeout activity >1m in the storage logs).

**Fix:** bound each read with a short timeout and move it off the request path into an async refresher that publishes an atomic snapshot, falling back to `DefaultCollapseSettings` on timeout/error. Operator edits still go live within one refresh interval — no per-call storage hit.

Validated end-to-end against the component-test harness: with this fix the storage apiserver no longer stalls and the previously-timing-out profile tests pass.

@coderabbitai review

entlein:
@coderabbitai — addressed in `e034f99`.

Good catch on the fallback regression — fixed. The refresher no longer overwrites the snapshot on a transient read error: on `err != nil` it now returns early and keeps the last known-good `CollapseConfiguration`, so an etcd blip / brief partition can't revert an operator-applied config to compiled defaults. Defaults apply only when the CR is genuinely absent.

One adjustment to the suggested diff: I kept the explicit `crd.Name == ""` guard rather than relying on `CollapseSettingsFromCRD(crd)` to default. With `IgnoreNotFound: true` an absent CR comes back as a non-nil zero struct, and `CollapseSettingsFromCRD` only defaults for a *nil* pointer — so dropping the guard would yield all-zero thresholds instead of defaults on the absent-CR path.

Also from the review:
- **Goroutine shutdown** — the refresher now takes an injectable stop channel (unexported core); the exported constructor passes `nil` (process-lifetime, as before). Wiring a real stop signal from the apiserver (`PostStartHook`/stop-channel) is left as the follow-up you flagged.
- **Async-refresh coverage** — replaced the old no-cache assertion (which contradicted this fix) with a test that drives the goroutine + ticker + `atomic.Pointer` path on a short interval, plus a transient-error-keeps-last-good test. The fake storage is now mutex-guarded so the suite is clean under `-race`.
- **Startup note** — documented that the one synchronous bounded refresh seeds the snapshot at construction and is never on the request path.

Provider tests pass under `-race`; component tests are green end-to-end against an image built from this commit.

entlein:
@coderabbitai review

pkg/registry/softwarecomposition/collapseconfiguration/strategy.go
```diff
@@ -0,0 +1,156 @@
...
+func (CollapseConfigurationStrategy) WarningsOnUpdate(_ context.Context, _, _ runtime.Object) []string {
+	return nil
+}
+
+// validateCollapseConfigurationSpec enforces the per-entry invariants and
+// rejects duplicate prefixes (which would silently produce a non-deterministic
+// longest-prefix-wins outcome at runtime).
+func validateCollapseConfigurationSpec(spec *softwarecomposition.CollapseConfigurationSpec, fp *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if spec.OpenDynamicThreshold < 0 {
```
matthyx, line 139: **Blocker: the global thresholds accept `0`, which collapses everything.**

`updateNodeStats` collapses a node once `node.Count > threshold` (`analyzer.go:334`), so an *effective* threshold of `0` collapses every node that has any children — i.e. every open/endpoint path becomes `⋯`.

This validator rejects only `< 0` for `OpenDynamicThreshold` / `EndpointDynamicThreshold`, while it requires each per-prefix `Threshold >= 1` (line 146). The inconsistency is dangerous: the v1beta1 JSON fields are **not** `omitempty`, so a CR that simply omits `openDynamicThreshold` decodes to `0`, and `CollapseSettingsFromCRD` passes it through verbatim (`collapse_config_from_crd.go:81`). The result is that applying a CR which sets only `collapseConfigs` silently zeroes the global thresholds and over-collapses all profiles in production.

Suggest requiring `>= 1` here too (consistent with the per-entry rule), or treating `0` as “use the compiled default” in `CollapseSettingsFromCRD`.

pkg/registry/file/dynamicpathdetector/collapse_config_from_crd.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	configs := make([]CollapseConfig, len(crd.Spec.CollapseConfigs))
+	for i, entry := range crd.Spec.CollapseConfigs {
+		configs[i] = CollapseConfig{
+			Prefix:    entry.Prefix,
+			Threshold: int(entry.Threshold),
+		}
+	}
+	return CollapseSettings{
+		OpenDynamicThreshold:     int(crd.Spec.OpenDynamicThreshold),
```
matthyx: **Related to the threshold-validation blocker.** This projection copies `Spec` values verbatim, with two consequences worth deciding explicitly:

1. A `0` global threshold is taken literally (see the `strategy.go` validation thread) → collapses everything.
2. `CollapseConfigs` is replaced **wholesale** — a CR with a single `collapseConfigs` entry drops all compiled-in per-prefix defaults (`/etc=100`, `/opt=50`, `/var/run=50`, …) rather than overlaying on top of them.

If wholesale-replace is intended, please document it on the CRD type and in `artifacts/collapseconfiguration-default-sample.yaml` so operators know a partial CR is not a partial override. If overlay was intended, merge with `DefaultCollapseSettings()` here and guard zero-valued thresholds.

matthyx (commented):
**Automated review of the current head (`e2346504`).**

Scope of this head is the `CollapseConfiguration` CRD + projection overlay, the cleanup user-managed skip, and the `networkmatch` differential test. Good news first:

- The three blockers from the 2026-05-27 review are resolved on this head: the `dynamicpathdetector` symbols (`NewPathAnalyzerWithConfigs`, `CollapseConfig`, `DefaultCollapseConfigs`, `FindConfigForPath`) exist on the branch, and the CRD is now wired into the live deflate path via `NewCRDCollapseSettingsProvider` → both processors.
- The two 2026-05-28 blockers (NetworkNeighbor `IPAddresses` not consumed in `networkpolicy.go` / `networkneighborhood_processor.go`) reference code that is **not in this PR's diff** anymore — `main...HEAD` touches only `networkmatch/match_differential_test.go` under `network*`. They appear resolved-by-rescope; worth confirming the `ipAddresses` feature landed (or is tracked) elsewhere.

One blocker remains, plus a few non-blocking notes.

---

### 🔴 Blocker — a partial / zero-valued CollapseConfiguration silently over-collapses every profile

See the inline comments on `strategy.go` and `collapse_config_from_crd.go`. In short: `updateNodeStats` collapses once `node.Count > threshold`, so an effective threshold of `0` turns **every** path into `⋯`. Validation allows the two global thresholds to be `0` (only `< 0` is rejected) while each per-prefix entry must be `>= 1`, and the v1beta1 JSON fields are not `omitempty` — so a CR that omits `openDynamicThreshold` (or only sets `collapseConfigs`) decodes to `0` and is passed through verbatim, silently nuking all opens/endpoints in production. Fix: require `>= 1` for the global thresholds too, or treat `0` as "use compiled default" in `CollapseSettingsFromCRD`.

### 🟡 Notes (non-blocking)

1. **Storage-instance asymmetry (apiserver.go).** The `collapseconfigurations` REST endpoint is registered as `ep(collapseconfiguration.NewREST)` → backed by the plain `storageImpl`, but the provider reads the CR from `applicationProfileStorageBackend` (the *with-collector* backend). It works today only because both point at the same `DefaultStorageRoot` on the same `OsFs`. Reading the CR through the same `storageImpl` the endpoint writes to would remove a latent footgun.
2. **Singleton name not enforced.** Only `CollapseConfiguration/default` is ever consulted; creating it under any other name is a silent no-op. Consider a `WarningsOnCreate`/`WarningsOnUpdate` when `metadata.name != "default"`.
3. **Breaking signature on an exported func.** `DeflateContainerProfileSpec` gained a `settings` parameter. If anything outside this repo imports it (e.g. node-agent), that's a compile break — please confirm there are no external callers or provide a wrapper.
4. **Truncated license header** in `pkg/registry/file/collapse_config_provider.go` (only the first 3 lines of the Apache header; the warranty paragraph other files carry is missing) — license-header linters may flag it.
5. **DCO check is currently failing** — independent merge gate; needs a `Signed-off-by` on all commits.

Nice work on the test coverage — the defensive-copy / nil-provider / live-update pins and the `networkmatch` differential oracle are thorough.

matthyx:
Re-checking the current `origin/main...HEAD` diff, I found one remaining code blocker:

- `pkg/registry/file/applicationprofile_processor.go:146`
- `pkg/registry/file/containerprofile_processor.go:826`

Both endpoint deflate paths call `AnalyzeEndpoints(..., NewPathAnalyzerWithConfigs(settings.EndpointDynamicThreshold, nil))`, which drops `settings.CollapseConfigs` entirely. The new `CollapseConfiguration` API documents `collapseConfigs` as per-prefix overrides for the dynamic-path-detectors open/endpoint collapse step, so applying the CR currently affects `opens` but not endpoint compaction.

Please pass `settings.CollapseConfigs` into the endpoint analyzer in both places and add a regression test that proves a CRD-supplied prefix override changes endpoint collapse output.

matthyx:
I found another blocker in the current diff:

- `pkg/registry/file/collapse_config_provider.go:36`

`collapseConfigurationKey()` builds the lookup key with `K8sKeysToPath("", "spdx.softwarecomposition.kubescape.io", "collapseconfigurations", "", "", name)`. Because both `cluster` and `namespace` are empty here, that helper expands to `/spdx.softwarecomposition.kubescape.io/collapseconfigurations//default` (note the extra `/`).

But `CollapseConfiguration` is registered as cluster-scoped (`pkg/registry/softwarecomposition/collapseconfiguration/strategy.go`), so the generic REST store persists it under the non-namespaced key `/<group>/<resource>/<name>` — i.e. `/spdx.softwarecomposition.kubescape.io/collapseconfigurations/default`.

Result: the provider reads the wrong path, misses the applied CR, and falls back to defaults every time. That makes applying `CollapseConfiguration/default` a no-op in practice.

Please build the cluster-scoped key without the empty namespace segment (or use the same key-generation path the REST store uses) and add an integration-style regression test that writes the CR through real storage and verifies the provider can read it back.

PR #328, opened 2026-06-01
Addon to PR 323: Moving Exec Args from NA to Storage

entlein:
currently https://github.com/kubescape/node-agent/pull/805 is built against my mirror of this branch, and the test results and benchmarks are at the bottom.

There are edits to outdated comments, the only `relevant` part of this PR is moving this function from NA to storage to avoid duplication of logic.

```
func argMatches(profileArg, runtimeArg string) bool {
	if profileArg == DynamicIdentifier {
		return true
	}
	if strings.Contains(profileArg, DynamicIdentifier) || strings.Contains(profileArg, WildcardIdentifier) {
		return CompareDynamic(profileArg, runtimeArg)
	}
	return profileArg == runtimeArg
}
```

## Summary by CodeRabbit

* **Refactor**
  * Improved exec-argument matching to more clearly separate constraint handling from wildcard/token matching, with stricter anchored behavior when required.

* **Tests**
  * Added coverage for embedded dynamic/wildcard matching, literal `*` staying literal, and boundary/adversarial cases to prevent unintended matches.
  * Added property-style and regression-style test suites plus a new benchmark for exec-arg matcher performance.

* **Documentation**
  * Removed upstream attribution from inline notes and clarified wildcard semantics and matching contracts.


matthyx (commented):
Reviewed the relocation of the exec-arg matcher into storage. The refactor reads clean, build is green, and `go test ./pkg/registry/file/dynamicpathdetector/...` passes. The doc-comment cleanup is fine. One blocker to resolve before merge, plus a nit.

### Blocker — `argMatches` reinterprets *literal* `*` / `⋯` in argv as wildcards

`compare_exec_args.go`:

```go
if strings.Contains(profileArg, DynamicIdentifier) || strings.Contains(profileArg, WildcardIdentifier) {
    return CompareDynamic(profileArg, runtimeArg)
}
```

The `strings.Contains` gate can't tell an analyzer-emitted wildcard token from a `*` that a process was genuinely invoked with. Any profile arg in which `*` (or `⋯`) lands on a `/`-delimited segment boundary now matches via `CompareDynamic` instead of literal equality, so the anchored match silently broadens. Confirmed against this branch:

- `CompareExecArgs(["/bin/*"], ["/bin/ls"])` → **true** (was literal-only: a profile that recorded the literal arg `/bin/*` now matches `/bin/<anything>`)
- `CompareExecArgs(["cmd","/var/log/*/a"], ["cmd","/var/log/X/a"])` → **true**

(An embedded `*` that is *not* a whole segment, e.g. `echo *`, still compares literally, so the divergence is specifically the segment-boundary `*` / trailing `/*` case.)

For a detection matcher this is a false-negative direction: a profiled exec entry matches more runtime argv vectors than before, so a malicious exec that happens to fall under a literal-glob arg won't trip R0040. The `argMatches` doc asserts "any `*` seen is embedded in a path token (the form the analyzer emits)" — that assumption holds for analyzer output but not for user-authored profiles or recorded literal-glob args, and there's no escaping (`\*` is noted as a v0.0.2 future).

Can you confirm whether this matches node-agent's current behavior (i.e. this is intended parity, not a new broadening), and whether literal-`*` exec args are considered out of scope? If it's intended, a one-line note on `argMatches` stating literal `*`/`⋯` in argv are deliberately treated as wildcards would close it. If not, the gate needs to distinguish a real token from a literal character.

### Nit
`analyzer.go` `hasPrefixAtBoundary`: the reflowed empty-prefix comment now breaks mid-sentence ("falls through to" / "pathPrefix[0] == '/'") and reads awkwardly — worth re-wrapping.

entlein:
New plan, we use a different non-overloaded character. I ll next update NA 805 and then ... we retest it all :)

matthyx (commented):
Re-reviewed at `35603ac`. The blocker is resolved — thanks for the thorough rework.

The exec matcher no longer overloads `*`. `⋯` (one arg / one segment) and the new `ExecArgsWildcard` `⋯⋯` (zero-or-more whole args) are dedicated sentinels, and `argMatches` now treats `*` as literal data. `matchExecArgSegments` does its own equal-count, literal-segment compare instead of delegating to `CompareDynamic`, so the file-path `*` semantics no longer leak into argv matching. Verified the previously over-broad cases are now literal:

- `CompareExecArgs(["/bin/*"], ["/bin/ls"])` → **false** (matches only literal `/bin/*`)
- `CompareExecArgs(["/plugins/*"], ["/plugins/evil.so"])` → **false**
- `CompareExecArgs([..,"/var/log/*/a"], [..,"/var/log/X/a"])` → **false**

…while the intended `⋯` path-token matching still works (`/usr/lib/postgresql/⋯/bin/postgres` matches the versioned runtime path and rejects an extra segment). Build is green and the full package test suite passes, including the new property / AP / glob-semantics tests. The comment-attribution nit is cleaned up too.

One coordination note (not blocking this PR): since `*` is now data and `⋯⋯` is the positional zero-or-more sentinel, node-agent (#805) must emit `⋯⋯` — not `*` — when it generates the "zero-or-more args" form, or learned profiles won't match. Worth a line in the NA PR description so the two land together.

LGTM on the storage side once #805 is aligned.

matthyx (approved):

entlein:
> 
> One coordination note (not blocking this PR): since `*` is now data and `⋯⋯` is the positional zero-or-more sentinel, node-agent (#805) must emit `⋯⋯` — not `*` — when it generates the "zero-or-more args" form, or learned profiles won't match. Worth a line in the NA PR description so the two land together.
> 


I dont think so , node-agent was never supposed to pre-aggregate the APs into `⋯⋯` , or did I misunderstand? The strange new character is only supposed to be externally supplied and now correctly used in comparison-statements, not created by the learnt profiles. 
I ll recheck, but that was kinda the whole reason not to use the `*` character

I m assuming the `⋯⋯` is from an out-of-band analysis, where a user explicitely states that some executable is intended to do have any argument.

PR #329, opened 2026-06-03
feat(validation): accept unsupported-schema status in ValidateStatusA…

harshitg927:
…nnotation

## Overview

`ValidateStatusAnnotation` is an allowlist that rejects any `kubescape.io/status`
value not in the known set. This adds the new `unsupported-schema` value so
objects carrying it pass validation instead of being rejected with "invalid value".

Context: kubevuln will write a stub VulnerabilityManifestSummary annotated
`kubescape.io/status: unsupported-schema` for images whose registry serves a
manifest schema the scanner can't read. While VulnerabilityManifestSummary
itself isn't currently routed through this validator, adding the value here keeps
the shared status vocabulary consistent and future-proofs any object type that
is validated by it.

## Changes

- `pkg/utils/validations.go`: add `helpers.UnsupportedSchema` to the
  `ValidateStatusAnnotation` switch.
- `pkg/utils/validations_test.go`: add a "valid status - unsupported schema" case.

## Related issues/PRs

Depends on kubescape/k8s-interface#155 (defines `helpers.UnsupportedSchema`).
Bump k8s-interface to the release that includes it, then drop the temporary
local `replace` before merge.

Part of: kubescape/kubevuln#369

## Checklist

- [x] Commits signed off (DCO)
- [x] Builds and tests pass locally (against the k8s-interface change)


matthyx:
@harshitg927 https://github.com/kubescape/k8s-interface/releases/tag/v0.0.214 is out

harshitg927:
Thanks @matthyx for the update, I will complete the work in kubevuln and storage pull request and let you know!

matthyx (approved):

PR #321, opened 2026-05-13
disable WatchList flag until we support it

matthyx:
## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled the WatchList feature in the file-based storage backend to prevent compatibility issues; the server startup now ensures WatchList remains off (ServerSideApply also remains disabled). A comment clarifies WatchList is not yet supported in file-based storage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/storage/pull/321)

Issue #320, opened 2026-05-13
Implement WatchList / sendInitialEvents protocol support


matthyx:
## Context

This is the follow-up implementation issue for #318, which reported that Rancher agents on k3s clusters enter a tight retry loop ("very short watch" error) when kubescape storage is installed alongside a cluster that has the `WatchList` feature gate enabled.

As an immediate mitigation, `WatchList=false` was added alongside the existing `ServerSideApply=false` in `pkg/cmd/server/start.go`. This prevents clients from using WatchList semantics against this server. Full protocol support is tracked here.

## Root Causes (from trace investigation)

Three converging gaps were found in the file-based storage backend:

### 1. Missing terminal BOOKMARK in `StorageImpl.Watch`
**File:** `pkg/registry/file/storage.go:388-401`

`StorageImpl.Watch` ignores `opts.SendInitialEvents` entirely. When a WatchList-aware client sends `watch?sendInitialEvents=true`, the server accepts the request but never emits the required terminal `BOOKMARK` event with annotation `k8s.io/initial-events-end: "true"`.

The standard Kubernetes `Cacher` (etcd wrapper in `k8s.io/apiserver/pkg/storage/cacher`) is the component that injects this bookmark — but it is not in the call chain here. The server wires `DryRunnableStorage{Storage: storageImpl}` directly, bypassing the `Cacher`. The HTTP watch handler (`endpoints/handlers/watch.go`) only relays BOOKMARK events from storage; it never injects them.

**Required change:** `StorageImpl.Watch` must check `opts.SendInitialEvents`, replay the current object list as `ADDED` events, then emit a `BOOKMARK{k8s.io/initial-events-end: "true"}` event before streaming live events.

### 2. `immutableStorage.Watch` returns a pre-closed channel
**File:** `pkg/registry/file/storage.go:1221-1222`

`immutableStorage.Watch` unconditionally returns `watch.NewEmptyWatch()` — a channel that is immediately closed. This affects the following resource types:
- `GeneratedNetworkPolicy` (`pkg/registry/file/generatednetworkpolicy.go:26`)
- Likely `ConfigurationScanSummary` and `VulnerabilitySummary` (also embed `immutableStorage`)

A pre-closed channel gives a WatchList reflector 0 items in < 1 second — the exact trigger for the "very short watch" tight retry loop. The old List+Watch reflector tolerated this (it interleaved a List between retries); the WatchList reflector does not.

[...]

PR #330, opened 2026-06-05
fix(server): make WatchList=false effective and stop pre-closed watch retry loops

### LINKED ISSUE #318: `WatchList` and `sendInitialEvents` support in storage
Does the kubescape storage backend support [streaming lists](https://kubernetes.io/docs/reference/using-api/api-concepts/#streaming-lists) (a.k.a. `WatchList` and `sendInitialEvents` REST query option)? I don't see any obvious handling of this in pkg/registry but I am far from an expert on building kubernetes aggregated api servers.

We recently ran into an issue on a k3s Rancher cluster that had the kubescape operator installed where the lack of  `WatchList`/`sendInitialEvents` REST option/streaming events appeared to cause the Rancher cluster agent to get into a state where the cluster was no longer accessible and responsive through the Rancher cluster management interface.

The rancher-agent container in the cattle-system namespace was producing many errors about a "very short watch" similar to the following:

`reflector.go:578] "Warning: watch ended with error" reflector="/root/.cache/go/modcache/k8s.io/client-go@v0.35.1/tools/cache/reflector.go:289" type="*summary.SummarizedObject" err="very short watch: /root/.cache/go/modcache/k8s.io/client-go@v0.35.1/tools/cache/reflector.go:289: Unexpected watch close - watch lasted less than a second and no items received"`

For reference this was observed with Rancher v2.13.2 and 2.14.1, and likely some preceding versions as well (we just didn't notice the issue until it caused an acute problem).

We ended up uninstalling the kubescape operator, which resolved the problem. We also were able to test the theory that `WatchList` support appeared to be the root cause by individually disabling the WatchList feature gate on the cluster.

Unfortunately Rancher appears to make heavy use of the streaming events/WatchList semantics on cluster resources to provide its functionality. Is this a known incompatibility with kubescape and/or is there a plan to support streaming events in the kubescape storage/apiserver layer?

matthyx:
## Summary

Fixes the Rancher cattle-agent tight retry loop reported in #318, which the mitigation in #321 could not resolve (confirmed by the reporter). Two independent server defects were found and fixed:

### 1. The `WatchList=false` feature-gate override was a silent no-op

`ComponentGlobalsRegistry.Set()` — the only bridge propagating the parsed `feature-gates` flag into `utilfeature.DefaultMutableFeatureGate` — ran **before** `flags.Set("feature-gates", ...)` populated the flag value, and was never re-invoked. The WatchList gate therefore silently stayed at its default (**enabled**), so the server accepted `watch?sendInitialEvents=true` requests (HTTP 200) it could never complete: the file-based storage has no Cacher and never emits the terminal `initial-events-end` BOOKMARK. This matches the reporter's `awaiting required bookmark event for initial events stream` log lines.

The calls are now reordered so the override is effective at request time. The apiserver then rejects `sendInitialEvents` watches **pre-stream (HTTP 422)**, and WatchList-enabled clients (client-go ≥ v0.35 ships `WatchListClient` Beta default-on; Rancher uses v0.35.1) fall back gracefully to legacy list+watch.

> **Note:** `ServerSideApply` is GA/non-gated in Kubernetes 1.35; #321's `ServerSideApply=false` token was always inert and is removed to prevent a boot-time `unrecognized feature gate` error now that flag ordering is fixed. There is **no** SSA behavior change in this PR. A negative test guards against reintroducing stale gate tokens.

### 2. Pre-closed watch channels caused "very short watch" tight loops

`watch.NewEmptyWatch()` returns a **pre-closed** channel. A reflector reading it observes 0 events in <1s → `VeryShortWatchError` → immediate re-watch → tight loop, in **both** legacy and WatchList modes — which is why disabling WatchList alone could never fully fix #318. Two sites were affected:

- namespaced watch keys in `StorageImpl.Watch`
- `immutableStorage.Watch` (ConfigurationScanSummary, VulnerabilitySummary, GeneratedNetworkPolicy)

Both now return `idleWatch`: a zero-goroutine `watch.Interface` that stays open and event-free until client disconnect or `Stop()` (closure driven by `context.AfterFunc` + `sync.Once`). Reflectors hold a quiet open connection instead of tight-looping. The real `watchDispatcher` path for cluster-scoped resources is untouched.

## Scope

This does **not** implement WatchList/`sendInitialEvents` semantics (#320) for the resources served by the real watch dispatcher — they still never emit `initial-events-end` bookmarks. Correctness relies on the now-effective `WatchList=false` producing the pre-stream 422 so clients never enter that path.

## Testing

- Regression test proving the gate is effective after `PersistentPreRunE` (fails on pre-fix code) + boot-safety + unrecognized-token guard + `skip` contract
- 422 pre-stream assertion at the exact handler decision point (`ValidateListOptions`)
- Idle-watch tests: stays open with zero events, closes on ctx-cancel / `Stop()` / double-`Stop()` / Stop-before-cancel; N=200 concurrent watches add ~0 goroutines (goleak-verified)
- Full suite green (`make test` — target added in this PR), `go vet` clean, `-race -count=3` clean
- Pending before merge: kind-cluster smoke with a client-go v0.35 reflector / Rancher to confirm fallback behavior end-to-end (a test image will be posted on #318)

Refs #318, #320

## Release Notes

* **New Features**
  * Improved watch behavior with idle watch support to prevent tight retry loops.
  * Enhanced feature-gate configuration handling.

* **Bug Fixes**
  * Watches now properly manage lifecycle and context cancellation.

* **Tests**
  * Added feature-gate validation and configuration tests.
  * Added idle watch behavior and performance tests.

* **Chores**
  * Updated build configuration with test runner.
  * Added testing dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,13 @@ TAG?=test
 IMAGE?=quay.io/kubescape/$(BINARY_NAME)
 
 
+.PHONY: build test docker-build docker-push
+
 build:
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o $(BINARY_NAME)
+
+test:
+	go test ./...
 
 docker-build:
 	docker buildx build --platform linux/amd64 -t $(IMAGE):$(TAG) --load -f $(DOCKERFILE_PATH) .
--- a/docs/features/watchlist-disabled.md
+++ b/docs/features/watchlist-disabled.md
@@ -1,0 +1,69 @@
+# WatchList disabled (streaming list / `sendInitialEvents`)
+
+## Summary
+
+The storage apiserver **disables the `WatchList` feature gate** (`feature-gates=WatchList=false`,
+set in `pkg/cmd/server/start.go`). As a result it **rejects `watch?sendInitialEvents=true`
+requests pre-stream with HTTP 422**, and WatchList-capable clients fall back to legacy
+`LIST` + `WATCH`.
+
+This is deliberate. The file/SQLite-backed storage has **no Cacher** and its watch path
+(`pkg/registry/file/watch.go`) only forwards *future* events — it never replays current
+state and **never emits the terminal `k8s.io/initial-events-end` BOOKMARK** that the
+WatchList (streaming-list) protocol requires. If the server accepted such a request, the
+client's reflector would block forever awaiting that bookmark.
+
+## Why it matters
+
+Kubernetes controllers that build **metadata informers** (`PartialObjectMetadata` LIST/WATCH)
+use WatchList by default once `WatchListClient` is on (client-go ≥ v0.35 ships it Beta /
+default-on; control planes from k8s ~1.32 enable it). Without this fix, those informers hang
+against the spdx aggregated API, which has two known, severe consequences:
+
+- **ResourceQuota controller stall.** The quota controller blocks `WaitForCacheSync` on
+  *every* discovered namespaced list+watch resource. One spdx monitor that never syncs
+  freezes **all** quota replenishment cluster-wide, so quotas stop decrementing and
+  eventually behave as if at their hard limit. (Upstream context:
+  [kubernetes/kubernetes#133737](https://github.com/kubernetes/kubernetes/issues/133737),
+  which names `spdx.softwarecomposition.kubescape.io/v1beta1`.)
+- **Rancher cattle-agent tight retry loop** ([#318](https://github.com/kubescape/storage/issues/318)):
+  reflectors awaiting the missing bookmark, plus pre-closed watch channels, drove a hot
+  re-watch loop.
+
+## How it works
+
+1. **Effective gate ordering.** `flags.Set("feature-gates", "WatchList=false")` must run
+   **before** `ComponentGlobalsRegistry.Set()` — that registry call is the only bridge that
+   propagates the parsed flag into `utilfeature.DefaultMutableFeatureGate`. If the order is
+   reversed (as it was before [#330](https://github.com/kubescape/storage/pull/330)), the
+   override is a silent no-op and the gate stays at its default (**enabled**). The apiserver's
+   watch handler then validates `ListOptions` against the now-disabled gate and returns 422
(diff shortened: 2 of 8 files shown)


matthyx:
## Review — no blockers found ✅

Checked out `fix/318-effective-feature-gates-idle-watch` and verified both fixes locally (Go 1.25): targeted tests pass, idle-watch tests clean under `-race -count=3`, `go vet` clean.

### Feature-gate ordering (`start.go`)
Reordering `flags.Set("feature-gates", "WatchList=false")` before `ComponentGlobalsRegistry.Set()` is the correct fix — `Set()` is the only bridge that copies the parsed flag into the live gates, so it has to run after the flag is populated. Dropping the inert `ServerSideApply=false` token (GA/non-gated since 1.35) is sound and well-guarded by the negative `unrecognized feature gate` test. The 422 assertion exercises the real `ValidateListOptions` decision point. 👍

### `idleWatch` (`watch.go`, `storage.go`)
Implementation is correct and the concurrency is safe:
- `close()` guarded by `sync.Once`; `Stop()` is idempotent and safe before/after/concurrent with ctx cancellation.
- `w.stop` is written once in `newIdleWatch` before the pointer escapes, so no race with the `AfterFunc` goroutine.
- Zero goroutines at steady state (verified by the goleak/`NumGoroutine` test and by `-race`).

Both pre-closed `NewEmptyWatch()` sites are converted; no others remain. No behavioral regression — namespaced and immutable watches never delivered events before either, they just no longer pre-close into a "very short watch" loop.

### Non-blocking nits (already raised by CodeRabbit, optional)
1. The test suite relies on process-global `DefaultMutableFeatureGate` state and source ordering (`TestPersistentPreRunERejectsUnknownGate` must stay last). It's documented, and CI doesn't use `-shuffle`, so it's fine today — but a save/restore (snapshot + `defer`) of the gate state would make it robust to `-shuffle=on` and future test additions.
2. The new `make test` target could match the README's `go test -v -failfast -count=1 ./...` to avoid cached-result false passes in CI.

Neither blocks merge. Pending kind/Rancher smoke (already noted in the PR description) is the right last gate before release.

PR #335, opened 2026-06-10
feat(collapse): rule-aware protection of sensitive open prefixes

matthyx:
## Problem

Profile deflation generalises high-cardinality open paths to keep stored profiles bounded (e.g. `/etc` with many files → `/etc/⋯`, or the whole tree → `/⋯/⋯`). That generalisation silently defeats anomaly rules such as **R0010 (unexpected `/etc/shadow` access)**: once `/etc` collapses to a wildcard, `ap.was_path_opened("/etc/shadow.evil")` matches it and the rule can never fire.

## Fix

Pin rule-declared sensitive prefixes (and their ancestors) to **literal** during collapse, so a never-seen sensitive path stays distinguishable from a generalised one. The analyzer skips collapsing any node on the ancestor chain of — or within the subtree of — a protected prefix.

- **`dynamicpathdetector`**: `NewPathAnalyzerWithConfigsAndProtection` + `OpenProtection` (`exact`/`prefix`/`suffix`/`contains`; `suffix`/`contains` are resolved against the observed open paths since they have no fixed location). `protectedNode` is **O(1)** for the common node via a precomputed ancestor set + top-dir fast-reject; cost never scales with the number of protected prefixes, and there is **zero overhead when no protection is configured**.
- **`DeflateContainerProfileSpec`** takes an `OpenProtection`; `OpenProtectionFromMatchers` converts `armotypes.OpenMatchers`. `config.Config.ProtectedOpenMatchers` feeds the in-cluster processor (populated by the operator from the rule library via `armotypes.UnionOpenProtection`); the backend feeds it from MongoDB rules.

## Tested

- `TestProtectedPrefixKeepsSensitiveDetectable` and `TestOpenProtectionExactSuffixContains`: with `/etc` far over the collapse threshold, never-seen sensitive paths (`/etc/shadow.evil`, `/etc/sudoers.bak`, `/etc/sudoers.d/99-evil`, `/etc/ssh/ssh_host_ed25519_key`, `/home/alice/.ssh/evil`) stay uncovered, while `/proc` still collapses (bloat control preserved). Existing `dynamicpathdetector` suite unaffected (zero-value `OpenProtection` = legacy behaviour).

## Performance

Deflation is per profile **PreSave** (consolidation), not per open event — off the per-event hot path. Benchmarked on a ~1,340-open profile:

| | per profile |
|---|---|
| No protection | ~0.9 ms |
| R0010 protection | ~1.3 ms (+0.4 ms) |

(The precomputed `protectedNode` keeps this at ~1.4× rather than the ~6.4× a naive per-prefix scan would cost.)

## Deps

- `armoapi-go` → pseudo-version `v0.0.719-0.20260610161827-5cfa1e630e5d` carrying `OpenMatchers` / `UnionOpenProtection` (armosec/armoapi-go#656). **Re-pin to the tagged release before merge.**

## Related
- armosec/armoapi-go#656 (`UnionOpenProtection` helper)
- kubescape/node-agent#834 (query-side schema dedup)

## Summary by CodeRabbit

* **New Features**
  * Configurable "protected" matchers and a runtime store to pin sensitive path prefixes so they are treated specially during path generalization; optional periodic reloader keeps them in sync.
  * Analysis now respects protected prefixes and caps protected-subtree growth to avoid unbounded memory use.
* **Tests**
  * Added regression tests covering protection rules and collapse behavior.
* **Dependencies**
  * Updated armoapi-go dependency version.

### DIFF
--- a/pkg/registry/file/containerprofile_processor.go
+++ b/pkg/registry/file/containerprofile_processor.go
@@ -45,18 +45,25 @@ type ContainerProfileProcessor struct {
 	MaxContainerProfileSize int
 	ContainerProfileStorage ContainerProfileStorage
 	ConsolidatedSlugChannel chan ConsolidatedSlugData
-	// OpenProtection is the union of sensitive open matchers (exact/prefix/
-	// suffix/contains) declared by active rules' profileDataRequired.opens.
-	// Matched prefixes (and their ancestors) are pinned to literal during
-	// deflation so rules like R0010 keep working. In-cluster this is populated
-	// from the rules CRD; the zero value preserves legacy collapse behaviour.
-	OpenProtection dynamicpathdetector.OpenProtection
+	// protection holds the active union of sensitive open matchers (exact/prefix/
+	// suffix/contains) declared by active rules' profileDataRequired.opens. Matched
+	// prefixes (and their ancestors) are pinned to literal during deflation so
+	// rules like R0010 keep working. It is read on every PreSave via Get and
+	// refreshed out-of-band by an OpenProtectionReloader (in cluster, fed from the
+	// operator-published ConfigMap); the zero value preserves legacy collapse
+	// behaviour.
+	protection *OpenProtectionStore
 }
 
-func NewContainerProfileProcessor(cfg config.Config, cleanupHandler *ResourcesCleanupHandler) *ContainerProfileProcessor {
+func NewContainerProfileProcessor(cfg config.Config, cleanupHandler *ResourcesCleanupHandler, protection *OpenProtectionStore) *ContainerProfileProcessor {
 	hostType := cfg.HostType
 	if hostType == "" {
 		hostType = armotypes.HostTypeKubernetes
+	}
+	if protection == nil {
+		// Seed from static config when no shared store is injected (e.g. backend
+		// callers and tests that don't run a reloader).
+		protection = NewOpenProtectionStore(cfg.ProtectedOpenMatchers)
 	}
 	return &ContainerProfileProcessor{
 		CleanupHandler:          cleanupHandler,
@@ -66,8 +73,14 @@ func NewContainerProfileProcessor(cfg config.Config, cleanupHandler *ResourcesCl
 		HostType:                hostType,
 		Interval:                30 * time.Second,
 		MaxContainerProfileSize: cfg.MaxApplicationProfileSize,
-		OpenProtection:          OpenProtectionFromMatchers(cfg.ProtectedOpenMatchers),
+		protection:              protection,
 	}
+}
+
+// ProtectionStore returns the processor's shared open-protection store so callers
+// (e.g. main wiring a reloader) can refresh it after construction.
+func (a *ContainerProfileProcessor) ProtectionStore() *OpenProtectionStore {
+	return a.protection
 }
 
 // OpenProtectionFromMatchers converts the shared armoapi-go matcher union into
@@ -200,7 +213,7 @@ func (a *ContainerProfileProcessor) PreSave(ctx context.Context, object runtime.
 	} else {
 		logger.L().Debug("ContainerProfileProcessor.PreSave - failed to get sbom name", loggerhelpers.Error(err), loggerhelpers.String("imageTag", profile.Spec.ImageTag), loggerhelpers.String("imageID", profile.Spec.ImageID))
 	}
-	profile.Spec = DeflateContainerProfileSpec(profile.Spec, sbomSet, a.OpenProtection)
+	profile.Spec = DeflateContainerProfileSpec(profile.Spec, sbomSet, a.protection.Get())
 	size += len(profile.Spec.Execs)
 	size += len(profile.Spec.Opens)
 	size += len(profile.Spec.Syscalls)
--- a/pkg/registry/file/containerprofile_processor.go
+++ b/pkg/registry/file/containerprofile_processor.go
@@ -45,6 +45,12 @@ type ContainerProfileProcessor struct {
 	MaxContainerProfileSize int
 	ContainerProfileStorage ContainerProfileStorage
 	ConsolidatedSlugChannel chan ConsolidatedSlugData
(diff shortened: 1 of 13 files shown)


pkg/registry/file/dynamicpathdetector/protection.go
```diff
@@ -0,0 +1,76 @@
...
+	if p.Empty() {
+		return nil
+	}
+	out := make([]string, 0, len(p.Exact)+len(p.Prefix))
+	out = append(out, p.Exact...)
+	out = append(out, p.Prefix...)
+	if len(p.Suffix) > 0 || len(p.Contains) > 0 {
+		for _, op := range openPaths {
+			if p.matchesUnanchored(op) {
+				out = append(out, op) // pin this open's ancestor chain
```
matthyx, line 60: **Blocker (partial coverage): Suffix/Contains protection is learning-dependent and does NOT protect never-seen paths — which is the whole point of the PR.**

Exact/Prefix pin *statically* (their ancestor chain is kept literal even if the sensitive file was never opened during learning). But Suffix/Contains are only pinned for paths that were *actually observed* in `openPaths`. If no sibling matching the matcher was opened during learning, and that directory is high-cardinality, the directory collapses to `/dir/⋯` and a first-ever access to a matching file is covered by the wildcard — i.e. the R0010-class false negative the PR sets out to fix still happens for these two matcher kinds.

Concrete: matcher `Suffix: ["_key"]`, and during learning `/etc/ssh` had 80 non-key files opened but **no** `*_key` file. `/etc/ssh` collapses to `/etc/ssh/⋯`, and a novel `/etc/ssh/ssh_host_ed25519_key` is then covered → rule can't fire. The test passes only because `ssh_host_rsa_key` *was* in the observed opens, which pins `/etc/ssh`; remove that one observed key and the guarantee disappears.

This asymmetry (static for exact/prefix, best-effort for suffix/contains) isn't surfaced anywhere a rule author would see it. Options: (a) document it loudly as a known limitation of suffix/contains matchers, (b) accept that suffix/contains can only ever be best-effort here and steer rule authors toward exact/prefix for must-detect paths. At minimum the PR description's "keeps a never-seen sensitive path distinguishable" should be scoped to exact/prefix.

pkg/registry/file/containerprofile_processor.go
```diff
@@ -807,8 +829,27 @@ func (a *ContainerProfileProcessor) getAggregatedData(ctx context.Context, key s
...
+func DeflateContainerProfileSpec(container softwarecomposition.ContainerProfileSpec, sbomSet mapset.Set[string], openProtection dynamicpathdetector.OpenProtection) softwarecomposition.ContainerProfileSpec {
+	var protectedPrefixes []string
+	if !openProtection.Empty() {
+		openPaths := make([]string, len(container.Opens))
+		for i := range container.Opens {
+			openPaths[i] = container.Opens[i].Path
+		}
+		protectedPrefixes = openProtection.ProtectedPrefixes(openPaths)
+	}
+	openAnalyzer := dynamicpathdetector.NewPathAnalyzerWithConfigsAndProtection(OpenDynamicThreshold, nil, protectedPrefixes)
```
matthyx, line 864: **Blocker to confirm (profile-size blowback): pinning a sensitive prefix makes its entire ancestor directory uncollapsible, which can flip the profile to `TooLarge`.**

Protecting e.g. `/etc/shadow` pins `/etc` to literal — so every distinct file ever opened under `/etc` now survives as its own entry instead of folding to `/etc/⋯`. On a chatty workload (or with several rule-protected top-level dirs), that growth feeds straight into the `size` accumulation in `PreSave`, and `size > MaxContainerProfileSize` sets `helpers.TooLarge`.

If a `TooLarge` profile is dropped from / not used by anomaly detection downstream, this is a self-defeating outcome: turning on protection for R0010 could disable detection for the whole container — strictly worse than the wildcard it was meant to fix. The Performance section measures *time* (+0.4 ms) but not the *size* delta, which is the dimension deflation exists to bound.

Please confirm what `TooLarge` does downstream, and add a worst-case size figure (e.g. `/etc` with N literals) next to the timing table. If `TooLarge` degrades detection, this needs a guard (cap pinned literals per protected dir, or fall back to collapse when a pinned subtree itself exceeds a budget).

matthyx (commented):
Reviewed the rule-aware collapse protection. The design is sound, it builds clean, and `go test ./pkg/registry/file/dynamicpathdetector/...` passes. The `protectedNode` O(1) fast-path and the zero-overhead-when-unconfigured guarantee both hold up.

The `armoapi-go` re-pin called out in the PR body is **already done** — `go.mod` is on the tagged `v0.0.719` (not the pseudo-version), the tag exists upstream, and armosec/armoapi-go#656 is merged. So that pre-merge item is cleared.

Two things I'd want resolved (or at least explicitly documented as accepted limitations) before merge — left inline. Neither is a compile/correctness bug; both are gaps between what the PR claims to guarantee and what it actually guarantees.

matthyx:
Re-reviewed `0841458`. Both blockers from my earlier review are addressed — builds clean, `dynamicpathdetector` tests pass (incl. the new `TestProtectedPrefixBudgetCollapse`):

1. **Suffix/Contains learning-dependence** — now documented as a `CRITICAL/KNOWN LIMITATION` in `protection.go` with explicit guidance to use Exact/Prefix where detection must be guaranteed. Good — the gap is real but now surfaced where rule authors will see it.

2. **Profile-size blowback** — `PinnedSubtreeBudget = 500` caps literals under a pinned node; past the budget it falls back to collapse rather than risk `TooLarge` (which clears the whole spec). Sound tradeoff: degrade one pathological dir instead of dropping the entire profile.

One non-blocking note: when the budget trips, protection silently flips off for that directory (the test confirms `/etc/shadow.evil` becomes covered again). Worth a `Debug` log when a pinned node hits the budget, so an operator can tell *why* an R0010-class rule stopped firing for a container. Not a merge blocker.

LGTM to merge.

PR #339, opened 2026-06-19
fix: paginate summary aggregation to avoid 500-item truncation (#337)

### LINKED ISSUE #337: GetList silently truncates to 500 items — ConfigurationScanSummary aggregation and prometheus-exporter report wrong per-namespace data on larger clusters
## Description

`pkg/registry/file/storage.go` applies a hard-coded default limit when a client lists objects without an explicit limit:

```go
// set default limit
if opts.Predicate.Limit == 0 {
    opts.Predicate.Limit = 500
}
```

A continuation token is returned, but two first-party consumers never use it:

1. **`ConfigurationScanSummaryStorage.GetList`** (`pkg/registry/file/configurationscansummarystorage.go`) builds the cluster-wide per-namespace summaries from a single `realStore.GetList` call on `workloadconfigurationscansummaries`. On a cluster with more than 500 stored summaries, the aggregation is computed **from the first 500 objects only**.
2. **kubescape/prometheus-exporter** (`api/api.go`) lists with `ResourceVersionFullSpec` and no limit, so its `kubescape_controls_total_namespace_*` / cluster metrics are computed from the same truncated page.

## Impact

On any cluster with >500 scanned objects (common in production — our mid-sized AKS cluster had ~640):

- namespaces that fall beyond the first page **silently disappear** from `ConfigurationScanSummary` list results and from all Prometheus metrics;
- per-namespace severity counters are wrong (we observed a namespace reporting `high: 0` via LIST while a direct GET of the same `ConfigurationScanSummary` returned `high: 91`);
- there is no warning anywhere that the data is partial — dashboards look healthy while being incorrect.

This is a security product reporting silently wrong security data, so flagging as high impact.

## Reproduction

1. Install kubescape-operator (chart 1.40.2, storage v0.0.274 or v0.0.281 — both behave the same) on a cluster with >500 workloads/objects covered by configuration scans.
2. `kubectl get --raw "/apis/spdx.softwarecomposition.kubescape.io/v1beta1/workloadconfigurationscansummaries?resourceVersion=fullSpec"` → exactly 500 items, `metadata.continue` set.
3. Compare `kubectl get configurationscansummary <ns> -o jsonpath='{.spec.severities}'` (correct) with the same namespace's entry in the cluster-wide LIST (wrong/missing).

## Suggested fix

- Consume the continuation token in `ConfigurationScanSummaryStorage.GetList` (and the vulnerability summary equivalent) so aggregations cover all objects; same in prometheus-exporter's client.
- At minimum, log a warning when a truncated list is used for aggregation.

## Workaround we use

Excluding namespaces from scanning to stay under 500 stored objects — not viable at fleet scale.

matthyx:
## Summary

Fixes #337 — on-the-fly summary aggregations silently truncated to the first 500 objects on larger clusters.

`ConfigurationScanSummaryStorage.GetList` and `VulnerabilitySummaryStorage.GetList` build cluster-wide, per-namespace summaries by listing the underlying `workloadconfigurationscansummaries` / `vulnerabilitymanifestsummaries`. The storage layer (`storage.go`) applies a hard-coded default `Limit = 500` when a client lists without an explicit limit and returns a continuation token — but these two aggregators issued a **single** `realStore.GetList` call and never consumed the token.

Result on clusters with >500 stored objects:
- namespaces beyond the first page **silently disappear** from the list results;
- per-namespace severity counters are wrong (a namespace can report `high: 0` via LIST while a direct GET returns the real count);
- nothing warns that the data is partial.

## Fix

Both `GetList` methods now follow the continuation token in a loop, accumulating every page before aggregating. The token is a monotonic SQLite `rowid` (`rowid > :cont ORDER BY rowid`), so paging is duplicate- and skip-free, and the loop terminates when the store returns an empty token (only set when a page is full).

## Tests

Added `..._GetList_Pagination` regression tests for both storages: create 500 objects in one namespace + 200 in a second (so the second namespace lives entirely on page 2), then assert both namespaces are present and all 700 objects are aggregated. Verified these fail against the pre-fix single-call code (1 namespace / count 500) and pass with the fix.

## Scope

- `kubescape/prometheus-exporter` (the other consumer named in the issue) is handled separately in kubescape/prometheus-exporter#52.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incomplete result retrieval for configuration and vulnerability summaries when query results exceed page limits.
  * Results now correctly aggregate across multiple pages.

* **Tests**
  * Added regression tests to verify pagination handling in configuration and vulnerability summary queries.

### DIFF
--- a/pkg/registry/file/configurationscansummarystorage.go
+++ b/pkg/registry/file/configurationscansummarystorage.go
@@ -97,9 +97,25 @@ func (s *ConfigurationScanSummaryStorage) GetList(ctx context.Context, key strin
 
 	workloadScanSummaryListObjPtr := &softwarecomposition.WorkloadConfigurationScanSummaryList{}
 
-	// ask for all workloadconfigurationscansummaries in the cluster
-	if err := s.realStore.GetList(ctx, "/spdx.softwarecomposition.kubescape.io/"+workloadConfigurationScanSummariesResource, opts, workloadScanSummaryListObjPtr); err != nil {
-		return err
+	// ask for all workloadconfigurationscansummaries in the cluster.
+	// The underlying store caps each page at a default limit and returns a
+	// continuation token, so we must page through every token to make sure the
+	// aggregation covers all objects and not just the first page (issue #337).
+	listOpts := opts
+	listOpts.Predicate.Continue = ""
+	for {
+		page := &softwarecomposition.WorkloadConfigurationScanSummaryList{}
+		if err := s.realStore.GetList(ctx, "/spdx.softwarecomposition.kubescape.io/"+workloadConfigurationScanSummariesResource, listOpts, page); err != nil {
+			return err
+		}
+		workloadScanSummaryListObjPtr.Items = append(workloadScanSummaryListObjPtr.Items, page.Items...)
+
+		// the store only sets a continuation token when the page was truncated;
+		// an empty token means we have read the last page.
+		listOpts.Predicate.Continue = page.Continue
+		if listOpts.Predicate.Continue == "" {
+			break
+		}
 	}
 
 	// generate a single configurationScanSummary for the cluster, with a configuration scan summary for each namespace
--- a/pkg/registry/file/configurationscansummarystorage_test.go
+++ b/pkg/registry/file/configurationscansummarystorage_test.go
@@ -2,6 +2,7 @@ package file
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


matthyx:
## Review — no blockers

I reviewed the pagination fix against the underlying store semantics and ran the tests. **No blocking issues found.**

### Correctness ✅
The loop matches the store's continuation contract exactly:
- `storage.go` sets a continuation token *only* when a page is full (`len(list) == limit`), and the token is the last SQLite `rowid`; `listMetadata`/`listMetadataKeys` query `rowid > :cont ORDER BY rowid`. So paging is strictly monotonic — duplicate-free, skip-free, and guaranteed to terminate.
- Resetting `listOpts.Predicate.Continue = ""` up front correctly discards any caller-supplied token (the aggregated `GetList` never emits one, so clients can't legitimately page it anyway).
- Accumulating `page.Items` across pages before aggregating is the right fix for the truncation.

### Tests ✅
Verified locally with Go 1.25:
- Both `..._GetList_Pagination` tests **pass on this branch**.
- Reverting only the two source files to `main` (keeping the new tests) makes both **fail** with exactly the truncation symptom — 1 namespace and `High: 500` instead of 2 namespaces / 700. So they are genuine regression tests, not tautologies. 👍

### Non-blocking note (no change required)
This trades "bounded-but-wrong" for "correct-but-unbounded": on a cluster with N summaries the aggregation now issues ⌈N/500⌉ sequential list pages and, in the `ResourceVersionFullSpec` path, one `get` per object, holding all N in memory before aggregating. That's inherent to a correct cluster-wide aggregation and fine for now, but if these endpoints ever get hit on very large clusters it'd be worth revisiting (e.g. aggregate incrementally per page rather than buffering all items). Not a blocker for this fix.

LGTM.

PR #348, opened 2026-07-22
Add CIDR-based collapsing for NetworkNeighbors

matthyx:
## Summary

- Adds CIDR-based collapsing for `NetworkNeighbor` entries, extending the existing path-collapsing pattern (`CollapseConfiguration` CRD, TTL-cached provider, `PreSave` hook) to the network side. Real-world case: node-agent bakes the literal destination IP into the `NetworkNeighbor` identifier hash, so every distinct external IP produces a separate entry — a real profile hit 8,687 such entries and was flagged `too-large`. See `docs/features/networkneighbors-collapsing.md` for the full mechanism, config knobs, and explicit accepted trade-offs.
- Fixes a pre-existing bug found while validating this feature: `compareNP` (test helper) compared `p1`'s Egress/Ingress against itself instead of `p2`, silently making every `compareNP`-based assertion in `TestGenerateNetworkPolicy` a no-op for years. Fixing it surfaced a real (harmless) latent bug — `mergeIngressRulesByPorts`/`mergeEgressRulesByPorts` could duplicate the same `IPBlock` peer 2-3x when the same IP appeared across multiple distinct `NetworkNeighbor` entries — now deduped.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes (no new warnings; one pre-existing unrelated warning in `callstack_test.go`)
- [x] `go test ./...` passes, including new unit tests for the collapse algorithm (idempotency, floor-bucketing, multi-bucket field replication), policy-generation tests for the new `IPAddresses`-consuming path, and the corrected `TestGenerateNetworkPolicy` fixtures
- [x] `hack/verify-codegen.sh` passes (CRD field additions regenerated)
- [x] Independently verified against the approved implementation plan (all 12 acceptance criteria checked with file/line evidence, idempotency traced rather than just test-passed)

Known follow-up (not fixed in this PR, flagged during investigation): `TestGenerateNetworkPolicy/real_duplicate_bug_test` is flaky due to `hash()` using `encoding/gob` over a struct containing a map — Go's randomized map iteration order occasionally produces different hashes for the same logical rule. Unrelated to this PR's changes.

AI-skills: oh-my-claudecode:plan,oh-my-claudecode:team | cmds: /oh-my-claudecode:deep-interview

## Summary by CodeRabbit

* **New Features**
  * Added CIDR-based collapsing for network neighbors with configurable thresholds for group sizing and minimum CIDR prefix.
  * Network policy generation now supports the plural `IPAddresses` field for ingress/egress peers (including CIDRs and wildcard).
* **Bug Fixes**
  * Avoids duplicate `IPBlock` peers during rule merging and improves determinism of generated policies.
* **Documentation**
  * Added a dedicated guide for CIDR-based collapsing and updated configuration samples/checklists.
* **Tests & Performance**
  * Expanded rule/collapse coverage with new unit tests and added benchmarks for the collapsing steps.

### DIFF
--- a/docs/features/networkneighbors-collapsing.md
+++ b/docs/features/networkneighbors-collapsing.md
@@ -8,7 +8,7 @@ The collapsing pass groups `NetworkNeighbor` entries that differ only by IP (sam
 
 Two new fields in the `CollapseConfiguration` CRD control this:
 - `NetworkIPGroupThreshold` (default 50): group size threshold above which IP collapsing is triggered.
-- `NetworkCIDRFloorBits` (default 16): minimum CIDR prefix length — no emitted block is ever broader than `/<floorBits>`.
+- `NetworkCIDRFloorBits` (default 24): minimum CIDR prefix length — no emitted block is ever broader than `/<floorBits>`.
 
 The pass is a **fixpoint**: running it twice on the same input is guaranteed to produce the same output, so collapsed entries remain stable across successive saves without oscillation or duplication.
 
@@ -29,8 +29,8 @@ Collapsing reduces entry count by orders of magnitude for traffic aimed at cloud
 
 **Aggregation**: If a group's count of aggregatable IPv4 host addresses exceeds `NetworkIPGroupThreshold`:
 1. Compute the smallest covering prefix (common leading bits across all hosts).
-2. If that prefix's length is at least `NetworkCIDRFloorBits` (e.g., `/16` or longer), emit it as-is.
-3. If it would be broader than `NetworkCIDRFloorBits`, split the group into floor-length buckets (e.g., `/16` buckets) and emit one CIDR entry per non-empty bucket.
+2. If that prefix's length is at least `NetworkCIDRFloorBits` (e.g., `/24` or longer), emit it as-is.
+3. If it would be broader than `NetworkCIDRFloorBits`, split the group into floor-length buckets (e.g., `/24` buckets) and emit one CIDR entry per non-empty bucket.
 
 **Output entries**: Each emitted CIDR entry carries:
 - The computed CIDR block(s) in `IPAddresses`.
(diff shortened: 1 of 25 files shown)


matthyx:
Good catch — confirmed. Added a third commit (2fee9177) fixing the actual root cause: \`hash()\` gob-encoded rules/policyRefs for dedup, but gob's map encoding follows Go's randomized map iteration order, so a rule containing a \`LabelSelector.MatchLabels\` map could hash differently across otherwise-identical calls. That's real product nondeterminism (generated NetworkPolicies weren't byte-stable run-to-run), not just a test artifact, as you noted.

Fix: switched \`hash()\` from \`encoding/gob\` to \`encoding/json\`, which sorts map keys deterministically.

Verification:
- Before: \`go test -run TestGenerateNetworkPolicy -count=40\` failed 3/40 on \`real_duplicate_bug_test\`.
- After: \`go test -run TestGenerateNetworkPolicy -count=100\` — 0/100 failures.
- Full \`go build ./...\` and relevant \`go test ./...\` still pass.

docs/features/networkneighbors-collapsing.md
```diff
@@ -0,0 +1,87 @@
...
+apiVersion: spdx.softwarecomposition.kubescape.io/v1beta1
+kind: CollapseConfiguration
+metadata:
+  name: default
+spec:
+  # ... existing fields (OpenDynamicThreshold, EndpointDynamicThreshold, CollapseConfigs) ...
+
+  # IP collapsing thresholds (optional; omit or set to 0 for defaults)
+  networkIPGroupThreshold: 50     # Collapse groups of 50+ hosts
+  networkCIDRFloorBits: 16        # No block narrower than /16
```
entlein: Isnt that a bit huge as the smallest allowed block size? /24 seems more reasonable

PR #351, opened 2026-07-29
chore: deprecate old profiletypes in favor of containerprofiles

entlein:
Removing all code relating to AP NN or ug

## Summary by CodeRabbit

* **New Features**
  * Network policies are generated from eligible container profiles and combined at the workload level.
  * Added execution, policy, and call-stack profile information.
  * Migration tooling supports container and seccomp profiles, including large numeric values.

* **Breaking Changes**
  * Application profile and network neighborhood resources and endpoints have been removed.

* **Bug Fixes**
  * Completed container profiles can no longer regress to Learning status.
  * Improved handling of unavailable profiles, policy generation, and migration errors.


cmd/migration/main.go
```diff
@@ -1,8 +1,8 @@
 package main
```
entlein, line 2: chicken, did we create this file /utility? 
Check if it is used anywhere in the codebase: if no remove it, if yes: we need tests

pkg/apis/softwarecomposition/v1beta1/networkpolicy/v2/networkpolicy.go
```diff
@@ -1,55 +0,0 @@
-package networkpolicy
-
-import (
-	sc "github.com/kubescape/storage/pkg/apis/softwarecomposition"
-	np "github.com/kubescape/storage/pkg/apis/softwarecomposition/networkpolicy/v2"
-	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
```
entlein, line 9: chicken, unless I m very much mistaken, we are not getting rid of the `NetworkPolicy` , only of the NN-profiles (and AP and ug-*)
reconfirm and check the entire codebase. If we keep it (my default assumption) make sure you restore all contracts, fixtures, tests, crds

pkg/registry/file/testdata/consolidate_golden/consolidated.golden.json
```diff
@@ -0,0 +1,185 @@
+{
```
entlein, line 1: chicken, whats this name "golden", that is not descriptive at all. same for window_*.json 
put in your memory: `descriptive short names without implementation details`

pkg/registry/file/cleanup.go
```diff
@@ -69,10 +69,11 @@ func initResourceToKindHandler(relevancyEnabled bool) map[string][]TypeCleanupHa
 		"workloadconfigurationscansummaries":  {deleteByWlid},
 	}
 
-	// only if relevancy is enabled, we need to delete application profiles with missing instanceId or wlid annotations
+	// only if relevancy is enabled, delete container profiles with missing
+	// instanceId or wlid annotations.
 	if relevancyEnabled {
 		logger.L().Debug("relevancy is enabled, adding additional cleanup handlers")
-		resourceKindToHandler["applicationprofiles"] = append(resourceKindToHandler["applicationprofiles"], deleteMissingInstanceIdAnnotation, deleteMissingWlidAnnotation)
+		resourceKindToHandler["containerprofiles"] = append(resourceKindToHandler["containerprofiles"], deleteMissingInstanceIdAnnotation, deleteMissingWlidAnnotation)
```
entlein, line 76: chicken, we need a test for this, cause cps have a different lifecycle than APs did. Lets have an exact contract here.

pkg/registry/file/containerprofile_aggregator_test.go
```diff
@@ -68,21 +69,6 @@ func (f *fakeStorage) GetTsContainerProfile(ctx context.Context, key string) (so
 func (f *fakeStorage) SaveContainerProfile(ctx context.Context, key string, profile *softwarecomposition.ContainerProfile) error {
 	return nil
 }
-func (f *fakeStorage) SaveMergedContainerProfile(ctx context.Context, observedKey string, profile *softwarecomposition.ContainerProfile) error {
-	return nil
-}
-func (f *fakeStorage) GetMergedContainerProfile(ctx context.Context, observedKey string) (softwarecomposition.ContainerProfile, error) {
-	return softwarecomposition.ContainerProfile{}, storage.NewKeyNotFoundError(observedKey, 0)
-}
```
entlein, line 76: chicken, why are we removing these?

entlein:
### Test-coverage gaps to close — ContainerProfile migration surface

Audit of the migration surface, prioritized by risk. Note two **"green test, wrong mechanism"** cases — a test exists but can't catch the real contract (the same failure shape as the GeneratedNetworkPolicy singular/plural bug that shipped in rc5k).

**HIGH**
- [ ] `cmd/migration/main.go` — **0%**. The one-shot gob→JSON migration decoder has no test/fixture; a field-decode bug silently corrupts every migrated profile. Golden test: encode a known legacy gob → decode → assert JSON round-trips each legacy `uint64` field (`LegacyArg`, `LegacySyscall.ErrnoRet`) + the spec shape; plus unsupported-type and corrupt-stream error cases.
- [ ] `pkg/registry/softwarecomposition/containerprofile/{etcd,strategy}.go` — **0%**. The CP REST strategy is copy-paste-named from SbomSyft (`SbomSyftStrategy`, `MatchWorkloadConfigurationScan`); a wrong `SelectableFields`/`GetAttrs` ships undetected. Strategy unit tests: `SelectableFields`/`GetAttrs` expose the CP namespace/name/workload labels; `NamespaceScoped()`; create/update hooks.
- [ ] **Completed-immutability guard** (`containerprofile_processor.go` PreSave, ~165-201). The revert logic is tested only against a fake whose `GetContainerProfileMetadataNoLock` *aliases* the locking read — so the whole reason the no-lock variant exists (avoiding the `GuaranteedUpdate` self-deadlock) is never exercised; swapping back to the locking call keeps tests green and deadlocks in production. Drive the guard through the **real** `GuaranteedUpdate` + sqlite pool; add the stored-`TooLarge` consolidated case.

**MED**
- [ ] `sqlite.go` ECS/Host `BuildContainerProfileKey`/`ParseContainerProfileKey`/`ECSKeysToPath`/`HostKeysToPath` — **0%** for non-K8s hosts (only K8s branch tested); a segment-order mismatch mis-scopes profiles. Round-trip table test per `HostType` (K8s new/legacy, ECS-EC2, ECS-Fargate, Host) + malformed-parts error branches.
- [ ] `generatednetworkpolicy.go` Get/GetList — untested: the `IsAvailable` exclusion (a non-available/Learning profile must be excluded — a regression that included it would pass), the knownservers-lookup failure, and the per-workload generate-error skip (`aggregateWorkloadPolicy` err → `continue` silently drops a workload).
- [ ] Consolidation dispatch `getAggregatedData` / `sendConsolidatedSlugToChannel` — **0%**; `listNamespaces` — **0%**; interleaved concurrent writes to the same CP key have no correctness test.

**LOW**
- [ ] `Is/NormalizeContainerProfileKind` — only one branch hit (the plural→singular key-matching lynchpin); `workloadPolicyName` no-kind-label skip path.
- [ ] networkpolicy/v2 + CIDR-collapse residual IP-parse/hash error branches (already 87–97%).

**Already solid (for reference):** the GNP workload-aggregation contract is now genuinely pinned (correct singular `containerprofile` key, cross-container union, one-policy-per-workload); networkpolicy/v2 87–97%; CIDR collapse 88–100%.

entlein:
chicken is claiming this is ready for review... chicken also tried to delete a repo.. twice.. today. I wouldnt trust it. 
@rabbit your turn, please review together with node-agent https://github.com/kubescape/node-agent/pull/864
and take into account https://github.com/k8sstormcenter/node-agent/actions/runs/30534641643
and the helm chart PR 
and the backend PR https://github.com/kubescape/backend/pull/55
and the synchronizer PR https://github.com/kubescape/synchronizer/pull/168

matthyx (changes requested):
Request changes based on an independent review — found a critical data-loss bug plus a few other gaps.

### Blocking

1. **Silent data loss in `cmd/migration/main.go:72`.** The embedded `metav1.LabelSelector` in `LegacyContainerProfileSpec` is tagged `json:"LabelSelector"`, but the destination `ContainerProfileSpec` (`pkg/apis/softwarecomposition/types.go:359`) embeds it untagged, so `encoding/json` promotes its fields instead of nesting them. The nested `"LabelSelector"` key in the tool's output never matches anything on unmarshal, `Spec.MatchLabels`/`Spec.MatchExpressions` come back empty, and `storage.go:712` writes that lossy object back over the original on disk. Downstream this can turn a per-workload NetworkPolicy's `podSelector` into `{}` (Kubernetes: selects every pod in the namespace). Fix: drop the tag so it matches the destination's field promotion (`metav1.LabelSelector` untagged, or `json:",inline"`).
2. **No round-trip test against the real destination type — please add one.** `cmd/migration/main_test.go` only gob-encodes fixtures with the same `Legacy*` structs the binary decodes into, then checks the tool's stdout as raw JSON. It never unmarshals that output into an actual `softwarecomposition.ContainerProfile`/`SeccompProfile`, which is the only thing that would have caught bug #1. Ask: add a test that (a) gob-encodes a fixture using an independent historical struct definition — not the same `Legacy*` types the binary uses — (b) runs the binary, and (c) `json.Unmarshal`s the output into the real destination type, asserting every field survives, in particular `Spec.MatchLabels`/`Spec.MatchExpressions`.
3. **Orphaned AP/NN data never reclaimed** (`pkg/registry/file/cleanup.go`) — the deprecated-resource cleanup map has no entries for `applicationprofiles`/`networkneighborhoods`, so old payload files and sqlite rows are kept forever on every upgraded cluster.
4. **Admission-time validation dropped** (`pkg/registry/softwarecomposition/containerprofile/strategy.go`) — `PrepareForUpdate`/`Validate`/`ValidateUpdate` are no-ops; the deleted `NetworkNeighborhoodStrategy` enforced completed-profile immutability and IP/DNS grammar validation with no replacement at the admission boundary.

### Worth fixing before/soon after merge, not necessarily blocking
- Out-of-`int64`-range legacy seccomp values (e.g. `SCMP_CMP_MASKED_EQ` masks) fail `json.Unmarshal` into `int64` fields — fails loudly rather than corrupting, but the profile becomes permanently unmigratable and gets retried on every read.
- Continuation token not reset in the new `generatednetworkpolicy.go` aggregation loop.
- `tests/integration-test-suite` is unmigrated and still asserts against `ApplicationProfile` — zero e2e coverage of the actual behavior change.

matthyx:
@entlein if you fix the migration bug, I'll merge

entlein:
I still need to review all the new CTs and manually rerun the demo and ... guess what: my favourite `opens gadget` fix wasnt complete, so I need to fix that FIRST, else node-agent main is buggy. 

I need either more of less amounts of ☕ 

I ll open the `opens-gadget fix` in NA asap.... 
Good news; the new Test_27 at least caught it

matthyx:
ok, so I should wait before releasing this? code is LGTM

entlein:
If you re willing to risk that the current CTs are an overlay of the various features (celnet, signatures(but off), cp-depreciateion, and the opens-gadget-fix) , it is green here https://github.com/k8sstormcenter/node-agent/actions/runs/31884625371

If you release it and I pull in your new main, I should be able to see very quickly if something broke and can fix it as a followup -> so if thats cool -> Please do go ahead ☺️

PR #354, opened 2026-08-05
fix: parse IPv6 hosts correctly in AnalyzeURL

ArneshBanerjee:
Part of #334 (AC-3)

AnalyzeURL adds http:// to any input that has no scheme, then hands it to url.Parse. For a bare IPv6 host this is wrong, because url.Parse needs the host inside brackets to tell it apart from the colons used as the port separator. Without brackets, something like ::1/health was parsed as host ":" and port "1", instead of host "::1" with no port at all.

This adds a step before the http:// prefix goes on: look at the authority part of the input, meaning everything before the first slash, and if it parses as an IPv6 address, wrap it in brackets. If the authority splits on its last colon into an IPv6 host and an all digit port, only the host part gets bracketed, so the port is kept. Anything that is not IPv6, or is already bracketed, is left alone.

The parsed host itself is discarded later in AnalyzeURL, only the port and path end up in the output, so the test asserts on that output. I used the same set of cases listed in the issue, IPv6 with and without a port, an IPv6 address with a trailing all digit group that looks like a port but isn't, and a few IPv4 and hostname cases to check nothing there changed.

Ran go test on the package and gofmt, both clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of scheme-less IPv6 addresses, including addresses with numeric ports.
  * Preserved correct behavior for bracketed IPv6 addresses, IPv4 addresses, and URLs with schemes.
  * Ensured IPv6 hosts are parsed with the expected ports and paths.

### DIFF
--- a/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
+++ b/pkg/registry/file/dynamicpathdetector/analyze_endpoints.go
@@ -3,6 +3,7 @@ package dynamicpathdetector
 import (
 	"encoding/json"
 	"fmt"
+	"net/netip"
 	"net/url"
 	"strings"
 
@@ -81,7 +82,7 @@ func ProcessEndpoint(endpoint *types.HTTPEndpoint, analyzer *PathAnalyzer, newEn
 
 func AnalyzeURL(urlString string, analyzer *PathAnalyzer) (string, error) {
 	if !strings.HasPrefix(urlString, "http://") && !strings.HasPrefix(urlString, "https://") {
-		urlString = "http://" + urlString
+		urlString = "http://" + addBracketsIfIPv6(urlString)
 	}
 
 	if err := isValidURL(urlString); err != nil {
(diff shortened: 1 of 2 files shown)


ArneshBanerjee:
@coderabbitai review

matthyx (approved):
Reviewed against AC-3 of #334. **No blockers — approving.** All nine table rows from the issue are implemented and asserted, and the fix is well scoped.

### What I verified locally (branch head `8845959`, base `72c61d4`)

- `go build ./...`, `go vet`, `gofmt -l` clean; `go test ./pkg/registry/file/...` green (5 packages).
- **Differential sweep, 1064 inputs** (19 authorities × 7 port forms × 8 path/query/fragment suffixes), base vs PR head:
  - 182 outputs changed — **every one of them an IPv6 authority**. Zero changes for hostnames, IPv4, already-bracketed hosts, or the canonical `:<port><path>` form, so the "must not change IPv4/hostname behavior" constraint holds.
  - **Zero success→error transitions** — no endpoint that used to parse is now dropped by `AnalyzeEndpoints`/`ProcessEndpoint`. 30 error→success improvements, e.g. `::1:abc/health` (a genuinely valid IPv6 address) went from `invalid port ":abc"` to `:/health`.
  - Zone IDs (`fe80::1%eth0`) still error before and after, only with a different message — matches the "declared unsupported" call in the issue.

### Non-blocking comments

**1. `?` and `#` don't terminate the authority** (same as CodeRabbit's inline note). `strings.IndexByte(urlString, '/')` misses query/fragment-only inputs, so `2001:db8::1?probe=1` still yields `:1.` — the exact bug this PR fixes, just with a `?` instead of a `/`. It is *not* a regression (base does the same), so it doesn't block, but the one-liner is free:

```go
if idx := strings.IndexAny(urlString, "/?#"); idx >= 0 {
```

I applied it locally: `2001:db8::1?probe=1` → `:.` and the whole package suite stays green.

**2. The `[host]:port` fallback is untested and currently unobservable.** Two checks:

- Mutation: deleting the entire `strings.LastIndex` branch *and* `isAllDigits` leaves the full package suite passing — no test pins it. (The case CodeRabbit suggests, `2001:db8::1:65535/x` → `:65535/x`, is worth adding, but note it passes with or without the branch, so it documents the contract rather than pinning the code.)
- Differential: for every input I could construct that reaches the branch — `2001:db8::1:65535/x`, `::1:65535/x`, `0:0:0:0:0:0:0:1:8080/x`, `fe80::abcd:12345/x`, `::ffff:192.168.1.1:8080/x`, `2001:db8:85a3::8a2e:370:7334:65000/api/v1` — the output is identical with and without it. `url.Parse` is lenient enough to split the last colon into the same port on the unbracketed string, and the host is discarded downstream.

So it's currently dead weight in output terms. I'd keep it (it makes the parse RFC-correct instead of leaning on `url.Parse` leniency, and it's what the issue specified), but a one-line comment saying it exists for parse correctness rather than for a different `:<port><path>` result would stop the next reader from hunting for the behavior difference.

**3. Heads-up, not a change request:** on IPv6 clusters, endpoints already recorded as `::1:8080/x` will re-key once from `:8080/x` to `:/x`, and `::1:0/x` stops being treated as the `:0` wildcard port. That is exactly the RFC 3986 call made in row 5 of the issue table, just worth knowing before it lands.

Also pre-existing and out of scope: a scheme-less input with no path (`::1`, `example.com`) produces the odd `:.` output from `AnalyzePath("")` — unchanged by this PR.

ArneshBanerjee:
Thanks!

